PR #113, opened 2026-04-15
Fix manage_wlan: both-platform comparison and assignment step

nowireless4u:
## Summary

Two fixes for issues found during testing:

### 1. SSID exists on both platforms — no comparison shown
When the AI called `manage_wlan_profile` and the SSID existed on both Mist and Central, the tool returned "found" for both but didn't include the actual configs. The workflow message now requires the AI to show a comparison table and ask the user to choose a source before proceeding. Both full configs are returned in the response.

### 2. Assignment loop
The workflow told the AI to "assign the WLAN profile to matching site collections" but there is no Central API tool for scope assignment. The AI got stuck in a loop trying to find a tool that doesn't exist. Fixed by changing Step 6 to tell the user to assign manually in the Central UI.

## Test plan
- [ ] CI passes
- [ ] Test with SSID on both platforms — AI shows comparison, asks which source
- [ ] Test Mist→Central sync — AI reports assignment to user, no loop

### DIFF
--- a/src/hpe_networking_mcp/platforms/manage_wlan.py
+++ b/src/hpe_networking_mcp/platforms/manage_wlan.py
@@ -53,15 +53,16 @@ opmode="WPA3_ENTERPRISE_CCM_128"
 **STEP 5** — Call `central_manage_wlan_profile(ssid="{ssid}", \
 action_type="create", payload=<mapped>)` to create the profile.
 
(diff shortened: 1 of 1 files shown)


nowireless4u:
Closing — will be replaced with a PR that includes the config-assignment tool for proper scope assignment.

PR #435, opened 2026-06-07
feat(skill): aos-migration target-architecture questionnaire (Stage 6.5) (v3.3.7.0)

nowireless4u:
## Summary

Reworks the `aos-migration` skill so the operator **decides** the target architecture via an explicit post-verdict questionnaire, instead of the skill auto-deriving-and-applying it. This is the operator-facing piece that ties the whole translation suite (clustering + wlan_ssid) together.

Flow is now: **source → readiness audit → verdict (Central-ready?) → questionnaire → translation plan.**

## What changed

**New Stage 6.5 — Target-architecture questionnaire** (runs after the readiness verdict, on operator `yes`, before the translation plan; `elicit()`-driven, adaptive):
- **Q1 — per-SSID forward mode:** `Bridged` / `Tunneled` / `Bridged-and-Tunneled` / `Hybrid` → drives `wlan_ssid` `target_mode`. Pre-filled from the detected `forward_mode`; the operator can pick anything (e.g. flip all to Bridged → controllerless).
- **Q2 — gateway topology** (only if any SSID survives on gateways): no-DMZ / campus+DMZ / DMZ-only → drives `gateway_cluster` `cluster_strategy` (`intent_site` / `intent_site`+`intent_manual` / `ha_only`). Skipped entirely if every SSID is Bridged.
- For `Bridged-and-Tunneled`: the bridge-vs-tunnel **scope split** (detected default, operator adjusts).

**Detection stays as the recommendation engine.** Stage 2 now *detects* per-SSID `forward_mode` + the SSID→cluster map; Stage 7 Step 4's cluster-mode derivation becomes the Stage 6.5 default. The "source says what *is*, operator says what they *want*" principle, in the skill.

**Removed the misaligned framing** throughout (per your request): "no configuration interview", "operator does not pick cluster mode", "auto-recommend ... override when reviewing". Those described the old model.

**Stage 9b wiring** — new preview subsections **2f** (`central:gateway_cluster`) and **2g** (`central:wlan_ssid`) that feed the Stage 6.5 decisions (`target_mode`, `cluster_strategy`, `gw_cluster_list`, bridge/tunnel scopes) into `central_translation_preview`.

## Scope
Skill-markdown only — no tool/code changes. The gate now routes `yes` → Stage 6.5 → Act II.

## Testing
- 90 skill-related tests pass (registration / frontmatter / live-detection catalog).
- Full suite: **1685 passed, 1 skipped**.

## Docs
- `CHANGELOG.md` 3.3.7.0; `docs/TOOLS.md` aos-migration entry (Stage 6.5 + replaced auto-derive description); `pyproject.toml` 3.3.6.0 → 3.3.7.0.

## Next
`wlan_ssid` inline bridge-mode AAA (the `aaa_prof` join) + radio/QoS, then the PII prereq before any write path.


nowireless4u:
## Code Review Summary

**Verdict: Changes Requested** — the docs-only change is mostly clear, and CI is green, but I found two workflow correctness issues that can make the skill block or crash the intended operator path.

### Critical

- **`src/hpe_networking_mcp/skills/aos-migration.md` around Stage 6 / Stage 6.5** — The questionnaire only runs after a non-`BLOCKED` verdict, but Stage 3 still evaluates rules using the *recommended* target mode before the operator can choose the actual target architecture.

  That creates a dead-end for the main use case this PR calls out: an operator may want to flip tunneled SSIDs to **Bridged** / controllerless, but if the pre-questionnaire verdict is `BLOCKED` because of gateway/tunnel-related findings, the Stage 6.5 questionnaire never appears. The operator never gets the chance to choose the architecture that would make those findings irrelevant.

  **Suggestion:** Either:
  - move the target-architecture questionnaire before the final verdict is computed,
  - compute the verdict with target-dependent findings separated from source-readiness findings, or
  - allow Stage 6.5 on `BLOCKED` when the blocking findings are target-architecture-dependent, then re-run Stage 3/Stage 6 against the chosen architecture.

### Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md` Stage 9b section 2f** — The gateway cluster preview loop processes every non-system/non-default `cluster_prof`, but Stage 6.5 Q2 only records decisions for clusters that survive into the chosen gateway topology. If some source clusters are unused, or if all SSIDs are bridged and Q2 is skipped, this snippet can hit `dec["cluster_strategy"]` with no decision present.

  **Suggestion:** Filter `cluster_records` to only clusters referenced by Stage 6.5 decisions / `gw_cluster_list`, and surface a clear `skip_reason` when a required decision is missing instead of indexing directly:

  ```python
  dec = decisions.get("per_cluster", {}).get(c["profile-name"])
  if not dec:
      cluster_previews.append({
          "source": c["profile-name"],
          "skip_reason": "no Stage 6.5 gateway-cluster decision for this source cluster",
      })
      continue
  ```

### Suggestions

- **Stage 6.5 wording / `decisions["per_ssid"]` shape** — The prose says “per-SSID,” but the example keys decisions by `"<vap-name>"`, and Stage 9b processes one `virtual_ap` at a time. That may be correct mechanically, but it is easy for an operator/implementer to read “SSID” as ESSID and accidentally collapse multiple VAP profiles that share the same SSID name across scopes.

  **Suggestion:** Make the identity explicit: either “per virtual AP / SSID instance” or include a stable key like `<source_scope>/<vap-name>` so duplicate VAP names or repeated ESSIDs cannot collide.

### Looks Good

- CI checks are passing: tests, lint/format, type check, Docker build, Bandit, dependency audit, and secret detection.
- The new Stage 6.5 intent is much clearer than the old “auto-derive then override later” wording.
- The docs correctly keep the workflow read-only and continue routing actual writes through preview/planning only.
- Changelog, tool docs, and package version were updated consistently.

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment** — skill-markdown/doc-only change. I don’t see security risk or executable code regression in the diff. The new Stage 6.5 flow is mostly clear and the changelog/docs/version bump line up. A couple of clarity/correctness gaps should be tightened so the playbook doesn’t send the operator into a ditch.

### Warnings
- **`src/hpe_networking_mcp/skills/aos-migration.md` — Stage 6.5 / Stage 9b gateway-cluster wiring**
  - Stage 6.5 says Q2 asks for gateway topology, and the `decisions["per_cluster"]` output includes `central_scope_id`, but the questionnaire text does not explicitly ask the operator to confirm the Central scope/name per surviving cluster.
  - Stage 9b then falls back to `scope_lookup["Global"]` if `central_scope_id` is missing:
    ```python
    "central_scope_id": dec.get("central_scope_id", scope_lookup["Global"])
    ```
  - That fallback could quietly preview gateway clusters at the wrong Central scope. Some clown left a trapdoor here: topology strategy and placement scope are separate decisions.
  - **Suggestion:** Make Stage 6.5 Q2 explicitly collect/confirm `central_scope_id` per surviving cluster, or treat missing scope as a `skip_reason` instead of defaulting to Global.

- **`src/hpe_networking_mcp/skills/aos-migration.md` — EMPTY-SOURCE handling**
  - Stage 6.5 says:
    > If verdict was EMPTY-SOURCE ... skip Stage 6.5 ... and proceed to Stage 7
  - But the gate logic above only defines BLOCKED / GO / PARTIAL prompt behavior. It does not define whether EMPTY-SOURCE gets a proceed prompt, stops, or is treated like PARTIAL.
  - **Suggestion:** Add explicit Act I gate behavior for `EMPTY-SOURCE`. If Stage 7 can still produce meaningful rows, say exactly what prompt appears. If not, stop cleanly.

### Suggestions
- **Stage 6.5 Q1 wording**
  - It says “Ask once per `virtual_ap` instance” and “for each translatable VAP,” which is good. Consider defining what makes a VAP “translatable” right there, especially for system/default/missing-SSID-profile cases. Stage 9b filters system/default later, but the questionnaire should match the preview loop so decisions don’t get requested for records that will be skipped.

- **Provisional finding re-score**
  - The split between verdict-gating findings and target-architecture-dependent provisional findings is solid. To make it harder for future edits to break, consider adding one short invariant near Stage 3/6:
    > Bucket 4 findings must never be included in the pre-questionnaire verdict count.
  - That’s the key rule. Put it where reviewers and future maintainers can’t miss it.

### Looks Good
- Changelog, docs entry, and `pyproject.toml` version bump are consistent for `3.3.7.0`.
- The shift from “auto-derived decision” to “detected recommendation → operator decision” is repeated clearly across the skill.
- Composite keying by `<source_scope>/<vap-name>` is the right move. Same ESSID/VAP names across scopes are exactly the kind of punk that causes bad migrations.
- Stage 9b guards missing Stage 6.5 decisions with `skip_reason` instead of indexing blindly. Now that’s solid work.

nowireless4u:
All three addressed in `4781a41` (skill-only).

**Critical — BLOCKED dead-ends the questionnaire.** Fixed by separating the verdict inputs. Stage 3 findings now split into four buckets; the pre-questionnaire verdict is computed from **buckets 2 + 3 only** (source-readiness + orchestration, target-architecture-independent). The per-target-mode rules (F3, F4, the Tunnel/Bridge/Mixed blocks — all gated on `target_mode_recommended`) are **bucket 4: PROVISIONAL** — scored against the *recommended* mode as a preview, never producing BLOCKED. So a gateway/tunnel REGRESSION can no longer suppress the gate, and the operator always reaches Stage 6.5 to pick Bridged/controllerless (which makes those findings moot). Added a **re-score step** in Stage 6.5: provisional findings are re-evaluated against the operator's *chosen* modes and folded into the Act II plan; if a chosen architecture still has a hard REGRESSION, the skill states it and offers to revisit. The gate text now says explicitly: "verdict computed from source-readiness + orchestration findings ONLY."

**Warning — 2f KeyError.** The gateway-cluster loop now does `dec = decisions.get("per_cluster", {}).get(name)` and, when absent (cluster unused / Q2 skipped because all SSIDs are Bridged), appends a `skip_reason` and continues — no direct `dec["cluster_strategy"]` indexing.

**Suggestion — per-SSID/VAP identity.** Q1 now explicitly asks **per `virtual_ap` instance, not per ESSID**, and decisions are keyed by the composite **`<source_scope>/<vap-name>`** (the dict renamed `per_ssid` → `per_vap`). The 2g loop builds that composite key and guards a missing decision. Same-ESSID VAPs across scopes can no longer collide.

Root cause (for my own accounting): the skill is markdown, so CI green validated the translations, not the skill's control flow — I added Stage 6.5 without tracing the unhappy paths (verdict=BLOCKED, missing decision, duplicate ESSID) or re-auditing how the new stage interacts with the untouched verdict-gating. Recorded as a standing check for future skill edits.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — found one correctness issue in the new Stage 6.5 decision model.

### ⚠️ Warnings

- **src/hpe_networking_mcp/skills/aos-migration.md:682 / 695 / 1307** — `decisions["per_cluster"]` is keyed only by `<cluster-name>`, and Stage 9b looks it up with `c["profile-name"]`.

  That can collide if the same `cluster_prof` name exists at multiple AOS hierarchy scopes. The skill already calls out this exact class of problem for VAPs and fixes it with the composite `<source_scope>/<vap-name>` key, but gateway clusters still use a bare profile name.

  **Why it matters:** a cluster decision could be applied to the wrong source-scope cluster, which means the generated `central:gateway_cluster` preview may use the wrong `cluster_strategy` or `central_scope_id`.

  **Suggestion:** key cluster decisions the same way VAP decisions are keyed, for example:

  ```python
  "<source_scope>/<cluster-name>"
  ```

  Then update the Stage 6.5 output shape and the Stage 9b 2f lookup to use that composite key.

### 💡 Suggestions

- **src/hpe_networking_mcp/skills/aos-migration.md:1342-1352** — Stage 6.5 says Q1 only asks for VAPs that resolve an `ssid_prof`, but the Stage 9b 2g preview filter only skips system/default VAPs. Missing `ssid_prof` cases will currently surface as the generic `"no Stage 6.5 forward-mode decision"` gap instead of the more useful missing-profile gap described earlier.

  Not a blocker, but tightening the pseudo-code to explicitly check the `ssid_prof` reference would make the playbook clearer.

### ✅ Looks Good

- The new Stage 6.5 flow is a solid improvement: source detection becomes a recommendation engine, while the operator makes the actual target-architecture decision.
- The split between verdict-gating findings and provisional target-architecture findings avoids locking the operator out of the questionnaire. Good guardrail.
- EMPTY-SOURCE behavior is now explicit.
- Stage 9b adds sensible skip paths instead of blindly indexing missing decisions.
- GitHub checks are green: tests, lint/format, type check, Bandit, dependency audit, secret detection, and Docker build all passed.

Fix that cluster key collision and this thing looks ready.

nowireless4u:
Thanks Casey — all four addressed in `15a76af` (skill-only). One of these was self-inflicted in my round-1 fix.

**Warning — `central_scope_id` defaulting to Global (the trapdoor).** This was my own round-1 regression: the 2f loop did `dec.get("central_scope_id", scope_lookup["Global"])`, which would silently preview a cluster at the wrong scope. Fixed both ends: Stage 6.5 **Q2 now explicitly collects `central_scope_id` per surviving cluster** ("strategy and placement are separate — collect both"), and 2f **skips with a `skip_reason`** when the scope is missing instead of defaulting to Global. Strategy (`cluster_strategy`) and placement (`central_scope_id`) are now clearly independent.

**Warning — EMPTY-SOURCE gate undefined.** Added explicit behavior in both the gate summary and the decision block: EMPTY-SOURCE emits its own proceed prompt; on `yes` it **skips Stage 6.5** (no SSIDs/clusters to decide) and runs Act II on defaults only; on `no` clean exit; `edit-context` not offered (no detected facts to correct).

**Suggestion — define "translatable" in Q1.** Q1 now states a VAP is translatable iff it matches the Stage 9b 2g filter (not `_flags.system`/`_flags.default`, resolves an `ssid_prof`). Missing-ssid-profile VAPs are flagged under Translation gaps, not asked about — and the 2g missing-decision guard catches any that slip through.

**Suggestion — verdict invariant.** Added a blockquoted **INVARIANT** right after the bucket list: *"Bucket-4 (target-architecture-dependent) findings must never be included in the pre-questionnaire verdict count."*

Since two rounds found issues, I did a **proactive adversarial sweep** this round rather than wait for a third: checked bucket-numbering consistency, every other `scope_lookup["Global"]` for the same trapdoor (the 2a–2e/2g ones are the documented Step-1 walker fallback with `<TBD:...>` placeholders, not silent defaults — only the Q2-owned cluster scope was the bug), and `per_vap`/`per_cluster` key consistency. CI green; 1685 tests pass.

nowireless4u:
Fixed in `679ae10` (skill-only) — you're right, and it was the same collision class I'd just fixed for VAPs and failed to apply to clusters.

**Warning (blocker) — cluster decision key collision.** `decisions["per_cluster"]` is now keyed by the composite **`<source_scope>/<cluster-name>`** in both the Stage 6.5 Q2 output shape and the Stage 9b 2f lookup, matching the VAP keying. A cluster decision can no longer be applied to a same-named cluster at a different source scope.

While fixing it I caught a dependency the composite keys *both* relied on but that was never established: the flat `stage1_<obj>_records` had no defined origin-scope. Added an explicit `_flatten()` at Stage 9b Step 2 that builds those lists from `config_by_scope` and **stamps each record with `_source_scope`** — without it, both `<source_scope>/<vap-name>` and `<source_scope>/<cluster-name>` would collapse to bare names and the collisions would silently return.

**Suggestion — 2g ssid_prof check.** The 2g loop now explicitly checks the `virtual_ap.ssid_prof` reference resolves in the collected `ssid_prof` set and emits the specific *missing-profile* gap, matching the Q1 "translatable" definition (rather than the generic "no decision" gap).

Accounting: three rounds, three finds — my sweeps were checking mechanical key-consistency, not the actual *collision-safety property* applied uniformly across analogous structures, and I'd relied on a `_source_scope` attribute nothing populated. Both lessons (apply a fix to every analogous structure; verify the data a fix depends on exists; never paper a missing required decision with a default) are now recorded. CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — the skill-only change is mostly coherent, and the targeted skill tests I ran passed, but I found two reliability issues in the new Stage 6.5/Stage 9b guidance that can steer the agent into the wrong rule path or a hard failure.

### ⚠️ Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md:700`** — The provisional finding re-score does not define how the new Stage 6.5 `target_mode` values map back to the existing rule gates.

  Stage 6.5 records `target_mode` as `bridged`, `tunneled`, `hybrid`, or `bridged_and_tunneled` (lines 663-668), but the rule gates still use the older values/sections: `target_mode_recommended == "tunnel"`, `"bridge"`, or `"mixed"` (lines 523-554). The new text says to re-evaluate the Tunnel / Bridge / Mixed blocks against each SSID’s chosen `target_mode`, but without a normalization table the agent can easily skip `hybrid` and `bridged_and_tunneled` or fail to apply both Bridge and Tunnel checks for dual-mode SSIDs.

  **Suggestion:** Add an explicit re-score mapping, for example:
  - `tunneled` → Tunnel block
  - `bridged` → Bridge block
  - `hybrid` → Mixed/Hybrid block, with any split-tunnel-specific rules
  - `bridged_and_tunneled` → both Tunnel and Bridge blocks, scoped to the operator’s `tunnel_scope_id` / `bridge_scope_id`
  
  Also clarify whether the old aggregate `mixed` concept still exists separately from per-SSID `hybrid` / `bridged_and_tunneled`.

- **`src/hpe_networking_mcp/skills/aos-migration.md:1391-1393`** — The WLAN SSID preview snippet can raise `KeyError` for incomplete dual-mode decisions.

  The surrounding text says a missing dual-mode scope should surface as a `skip_reason`, but the snippet directly indexes:

  ```python
  rt["bridge_scope_id"] = dec["bridge_scope_id"]
  rt["tunnel_scope_id"] = dec["tunnel_scope_id"]
  ```

  If Stage 6.5 recorded `target_mode == "bridged_and_tunneled"` but one scope is missing, this blows up instead of adding the promised translation gap.

  **Suggestion:** Guard both scope values before calling `central_translation_preview`, e.g. skip with a clear reason when either `bridge_scope_id` or `tunnel_scope_id` is missing.

### ✅ Looks Good

- The PR correctly removes the old “auto-derive-and-apply” framing and replaces it with an explicit post-verdict operator decision point.
- Composite keys for VAP and cluster decisions are a solid fix for same-name/same-ESSID cross-scope collisions.
- The EMPTY-SOURCE flow is clearer now and avoids asking pointless architecture questions.
- Stage 9b now includes gateway cluster and WLAN SSID preview wiring, which matches the stated Stage 6.5 goal.

### Tests run

```text
uv run --frozen pytest tests/unit/test_skills.py tests/unit/test_skill_tool_references.py tests/unit/test_skill_snippet_sandbox_compat.py tests/unit/test_skill_aos8_live_detection.py tests/unit/test_skill_cross_platform_rf_check.py
```

Result:

```text
86 passed, 5 warnings
```

Solid direction, but those two punks need cleaning up before merge.

nowireless4u:
Both fixed in `984e5d9` (skill-only).

**Warning — re-score vocabulary mismatch.** Added an explicit normalization table to the re-score step: `tunneled`→Tunnel block, `bridged`→Bridge block, `hybrid`→Mixed block (+ split-tunnel-specific), `bridged_and_tunneled`→**both** Bridge (scoped to `bridge_scope_id`) and Tunnel (scoped to `tunnel_scope_id`). Also clarified that the old aggregate `mixed` only existed pre-questionnaire to gate the provisional preview — post-questionnaire there's no aggregate mode; re-scoring is strictly per-SSID by the table, so a deployment with a mix of bridged + tunneled SSIDs is just each SSID under its own block (nothing is "mixed" unless it's `hybrid`/split-tunnel).

**Warning — dual-mode scope KeyError.** The 2g snippet now guards both `bridge_scope_id` and `tunnel_scope_id` before indexing and, if either is missing on a `bridged_and_tunneled` decision, appends the promised `skip_reason` and continues — no KeyError.

I also re-swept every `dec[...]` direct-index: all are now either behind a guard or a core field guaranteed by the `if not dec` check, and the `target_mode_recommended == "tunnel"/"bridge"/"mixed"` gates are the pre-questionnaire Stage 3 checks now bridged to the 4-value Stage 6.5 vocab by the re-score table. CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — the docs-only change is headed in the right direction, but I found a couple workflow correctness gaps that can send the skill down the wrong path or produce misleading previews. Some clown left a few doors cracked open.

### ⚠️ Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md:649` / `657-661` / `352-379` — Stage 6.5 is mandatory for GO/PARTIAL, but the offline upload/paste path has no VAP/SSID/cluster inventory.**  
  Stage 1-ALT explicitly only seeds parsed `netdst`, `acl_sess`, and `role`, then marks the rest inconclusive. But the gate now says GO/PARTIAL `yes` always proceeds to Stage 6.5, and Q1 is “always” per `virtual_ap`. On upload/paste runs, there may be no `virtual_ap` records at all because that parser path does not collect them.  
  **Suggestion:** Add explicit behavior for “no translatable VAPs available” / offline-source coverage, e.g. skip Stage 6.5 and proceed with parsed-object translations only, or emit a clear “Stage 6.5 requires API inventory / pasted effective config with VAP + SSID profile coverage” gap.

- **`src/hpe_networking_mcp/skills/aos-migration.md:997` vs `1323-1418` — Stage 9b’s minimal collection path was not updated for the new gateway/WLAN preview sections.**  
  Stage 9b says if the operator jumps straight to preview, collect only `role`, `acl_sess`, `vlan_id`, `vlan_name`, `vlan_name_id`, `netdst`, and `netdst6`. The new 2f/2g sections require `cluster_prof`, `virtual_ap`, `ssid_prof`, source-scope stamping, SSID→cluster decisions, and Stage 6.5 answers. That mismatch will produce empty/undefined preview inputs or misleading “no decision” gaps.  
  **Suggestion:** Update the Stage 9b preconditions/minimal collection list to include the new WLAN/gateway dependencies, or state that 2f/2g only run after full Act I + Stage 6.5.

- **`src/hpe_networking_mcp/skills/aos-migration.md:1395` — WLAN preview silently defaults `central_scope_id` to Global.**  
  The new cluster preview is careful not to silently default cluster placement to Global, but the WLAN preview hardcodes `scope_lookup["Global"]` with a comment saying “or the Stage-7 AP scope.” That can produce a deterministic-looking preview at the wrong Central scope, especially after the operator confirmed hierarchy placement in Stage 7.  
  **Suggestion:** Resolve the WLAN target scope from the Stage 7 source-scope → Central-scope mapping, or use a `<TBD:...>` placeholder / skip reason when the scope is not known. Don’t make Global look authoritative unless that’s actually the chosen target.

### 💡 Suggestions

- **`src/hpe_networking_mcp/skills/aos-migration.md:1422` — Preview report still says “Combine the five `result` dicts.”**  
  With gateway clusters and WLAN SSIDs added, this is no longer five.  
  **Suggestion:** Update to “all preview result dicts” or enumerate the current set.

- **`src/hpe_networking_mcp/skills/aos-migration.md:1001-1010` — Translation catalog table doesn’t list the new `central:gateway_cluster` and `central:wlan_ssid` previews.**  
  The new sections document them later, but the “Translations shipped today” table still stops at policy/net-group.  
  **Suggestion:** Add both rows so operators see the full preview surface in one place.

### ✅ Looks Good

- The Stage 6 verdict separation is solid: source-readiness/orchestration findings gate the verdict, while target-architecture findings stay provisional until the operator chooses the target. That fixes the old dead-end where a recommended target mode could block the questionnaire.
- Composite keys for `<source_scope>/<vap-name>` and `<source_scope>/<cluster-name>` are the right move. Same-name objects across AOS hierarchy scopes are a real collision risk.
- The new skip-reason handling in 2f/2g is practical and operator-friendly. Better to surface missing decisions than throw a KeyError or fake a preview.
- Changelog, tool docs, and version bump are aligned with the feature scope.
- No obvious secrets, conflict markers, or debug leftovers in the diff.

CI was still in progress when reviewed, so I’m not treating checks as final yet.

nowireless4u:
All five addressed in `30dc91d` (skill-only). #3 was the same scope-trapdoor class as the cluster one — and I'd explicitly rationalized 2g's `Global` default as "fine" in my own sweep last round. It wasn't.

**Warning — offline path has no VAP/SSID/cluster inventory.** Added an inventory-coverage precondition to Stage 6.5: if there are no translatable VAPs (offline upload/paste collects only `netdst`/`acl_sess`/`role`, or the walk found none), **skip Stage 6.5 entirely** and proceed to Act II for the parsed object classes, emitting a clear gap (*"requires `virtual_ap`/`ssid_prof`/`cluster_prof` — re-run on the API path or paste full effective-config"*). Never asks Q1/Q2 against an empty VAP set.

**Warning — Stage 9b minimal-collection mismatch.** Documented that the minimal jump-to-preview path covers **2a–2e only**; **2f/2g require** `cluster_prof`/`virtual_ap`/`ssid_prof` with `_source_scope` stamping **plus** the Stage 6.5 decisions, so they run only after a full Act I + Stage 6.5.

**Warning — 2g silent Global scope.** Now resolves the VAP's `_source_scope` via the Stage 7 mapping with a `<TBD:...>` fallback, never a silent authoritative `Global`. Added a **Step-1 scope-resolution principle** that applies to *every* 2a–2g snippet (pass the Stage-7-resolved scope; `<TBD:...>` when unknown; Global only when Global is genuinely the target) — so the rule is uniform instead of patched in one spot.

**Suggestions.** "Combine the five `result` dicts" → "all preview result dicts"; added `central:gateway_cluster` + `central:wlan_ssid` rows to the translation catalog (with an AAA-chain note) and refreshed the stale `v3.0.1.14` stamp to `v3.3.7.0`.

CI green; 1685 tests pass (one skill-tool-reference test caught a `central_name` pseudo-var I'd named — renamed). I'll wait for your next pass before calling it done.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — one workflow/control-flow issue needs tightening before merge.

### 🔴 Critical
- None found.

### ⚠️ Warnings
- **src/hpe_networking_mcp/skills/aos-migration.md:649 / :659 / :674** — Stage 6.5 now runs **before Stage 7**, but the new questionnaire says its gateway-topology defaults come from **Stage 7 Step 4**. That leaves the playbook ordering inconsistent: a user following the skill reaches Q2 before the Stage 7 cluster-mode derivation has actually run.

  **Why it matters:** Q2 is supposed to present detected topology defaults and collect `cluster_strategy` / `central_scope_id`. If the Stage 7 Step 4 derivation has not happened yet, the skill may either lack the recommendation data or force the AI/operator to improvise. That is exactly the kind of hidden train wreck this PR is trying to avoid.

  **Suggestion:** Move the cluster-mode recommendation derivation into Stage 2 / Stage 6.5 pre-work, or reorder the flow so Stage 7’s derivation runs before Stage 6.5 asks Q2. The text should make the data dependency explicit, e.g. “run the Stage 7 Step 4 derivation logic as part of Stage 2 detection” if that’s the intended model.

### 💡 Suggestions
- Consider renaming references from “Stage 7 Step 4” to something stage-neutral like “cluster-mode recommendation derivation” if the logic is now consumed before Stage 7. Cleaner playbook, fewer chances some future clown puts the dependency back in the wrong spot.

### ✅ Looks Good
- The Stage 6.5 design is much clearer than the old auto-derive-and-apply model.
- Composite keys for `<source_scope>/<vap-name>` and `<source_scope>/<cluster-name>` are solid — avoids cross-scope collisions.
- Good guards around missing Stage 6.5 decisions and missing `central_scope_id`; skipping with `skip_reason` is the right move.
- The EMPTY-SOURCE behavior is explicit now.
- Security-wise, no secrets or write-path behavior introduced in this diff.

### Checks observed
- `gh pr diff 435 --repo nowireless4u/hpe-networking-mcp` reviewed.
- GitHub checks at review time: security/lint/docker passed; **Tests** and **Type Check** were still pending.

nowireless4u:
Fixed in `2ef02a2` (skill-only). Real ordering bug — Stage 6.5 runs before Stage 7, so Q2's defaults can't depend on Stage 7 having run.

**Warning — Stage 6.5/Stage 7 ordering dependency.** Resolved by making the cluster-mode derivation a **Stage 2 detection** activity (it reads only Stage 1 data — `config_by_scope` / `ap_database` / `ap_multizone_prof` — all available in Act I):
- Stage 7 Step 4 is reframed as the **algorithm definition** that **runs during Stage 2 detection** and is **applied** (not re-derived) in Stage 7; added an explicit ordering note there. By the time Stage 7 runs, the operator has already confirmed/overridden the recommendation in Stage 6.5, and Stage 7 applies that.
- Stage 2 detection table now says "run the cluster-mode recommendation derivation here."
- Stage 6.5 Q2 + intro, the line-191 detection bullet, and the Stage 8 matrix now reference "Stage 2 detection / the cluster-mode recommendation derivation" instead of the not-yet-run Stage 7.

Per your suggestion I also moved toward stage-neutral naming ("cluster-mode recommendation derivation") wherever the logic is consumed before Stage 7. CI green; 1685 tests pass.

Waiting for your next pass before calling it done.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — one workflow bug needs tightening before merge.

### Critical
- None.

### Warnings
- **src/hpe_networking_mcp/skills/aos-migration.md:580-581** — The new design says target-architecture-dependent findings must not gate the pre-questionnaire verdict, but Stage 4 still has target-dependent orchestration checks that can fire as `REGRESSION` before Stage 6.5:
  - `A4` is explicitly gated on `target_mode_recommended in {bridge, mixed}`.
  - `A5` is gated on `cluster_profile_present`, but if the operator chooses all SSIDs as Bridged/controllerless in Stage 6.5, those gateway/VRRP NAD requirements may no longer apply.

  That creates the same dead-end this PR is trying to remove: a recommended bridge/mixed/gateway design could block the operator before they get the chance to choose a different target architecture.

  **Suggestion:** Move `A4` and the gateway-specific parts of `A5` into the same provisional/re-score flow as the per-target-mode findings, or explicitly state they are excluded from the pre-questionnaire verdict and re-evaluated after Stage 6.5 based on chosen `target_mode` / surviving gateway clusters.

### Suggestions
- **src/hpe_networking_mcp/skills/aos-migration.md:674-684** — For the “Campus gateways + DMZ clusters” option, consider spelling out how each surviving cluster gets classified as `intent_site` vs `intent_manual` when recording `decisions["per_cluster"]`. The current text says the topology option drives both strategies and then says each cluster gets a strategy, but the per-cluster assignment step could be clearer for the operator/model.

### Looks Good
- Composite keys for VAPs and clusters avoid the same-ESSID / same-profile-name cross-scope collision problem. Solid fix.
- Stage 9b preview loops guard missing decisions with `skip_reason` instead of indexing straight into partial data. Good defensive play.
- EMPTY-SOURCE and offline parser behavior are now explicit instead of leaving the workflow to guess.
- CI is green: tests, type check, lint/format, Docker build, Bandit, dependency audit, and secret detection all passed.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — one correctness issue in the Stage 6.5 → Stage 9b wiring.

### ⚠️ Warnings

- **src/hpe_networking_mcp/skills/aos-migration.md:695-699, 1424** — `gw_cluster_list` is documented/stored as composite cluster key strings like `"<source_scope>/<cluster-name>"`, then passed straight into `central_translation_preview`.

  The shipped `central:wlan_ssid` translation expects `runtime_values["gw_cluster_list"]` to already be the Central overlay payload shape, not source decision keys. The target schema documents this as:

  `[{cluster, cluster-type, tunnel-type, cluster-redundancy-type, cluster-scope-id}]`

  If the skill follows the new docs as written, the overlay-wlan preview will emit a `gw-cluster-list` containing source-key strings instead of Central cluster binding objects. That makes the generated plan look valid while producing the wrong payload shape — some clown left a mismatch between the questionnaire decision model and the engine contract.

  **Suggestion:** keep the composite keys inside `decisions["per_vap"]` if they’re useful for collision-safe lookup, but before calling `central_translation_preview`, resolve them through `decisions["per_cluster"]` / Stage 9b cluster preview output into the actual Central `gw_cluster_list` object list expected by `central:wlan_ssid`. The 2g snippet should show that conversion explicitly, or change Stage 6.5 to collect/store `gw_cluster_list` in the final Central payload shape.

### ✅ Looks good

- The source-readiness vs target-architecture-dependent verdict split is clear and fixes the old “recommended target blocks the questionnaire” trap.
- Composite `<source_scope>/<name>` keys for VAPs and clusters are the right move to avoid cross-scope collisions.
- EMPTY-SOURCE behavior is now explicit.
- Stage 9b guards missing decisions/scope data with `skip_reason` instead of indexing into missing keys. Solid work.
- Version/docs/changelog updates are aligned.

### Checks run

- `uv run --frozen pytest tests/unit/test_skills.py tests/unit/test_skill_aos8_live_detection.py tests/unit/test_skill_tool_references.py`
- Result: **60 passed**, 5 third-party `pyclearpass` syntax warnings.

nowireless4u:
Took a step back and did a **holistic pass** instead of patching the latest finding in isolation — read the whole skill, enumerated every issue *class*, and ran an independent adversarial review over the full file. That caught the thing I kept doing: I'd fix a class in Stage 3 but leave the *same class* in the downstream report/decision artifacts. Fixed in `eae11a0`:

**Verdict/bucket (the recurring dead-end), now fixed at EVERY site, not just Stage 3:**
- **Decision matrix** said "only F1–F10 + O1–O3 trigger BLOCKED" — but F3/F4 are bucket-4 (provisional) and A1–A3 also gate. Rewritten: verdict-gating = F1/F2/F5–F10 + O1–O3 + A1–A3; bucket-4 (F3, F4, T/B/M, A4, A5) **never** BLOCKs.
- **Act I REGRESSION template** listed bucket-4 findings (F3, F4, T3, T4, B2, B4, B5, M2) under "must fix before migration" — moved them to a separate **"Provisional (target-dependent) — re-scored after Stage 6.5, NOT verdict-gating"** section.
- bucket-4 definition + re-score step now name **A4/A5** alongside F3/F4/T/B/M.
- **F3/F4 are deployment-level** — defined their post-questionnaire trigger explicitly (any SSID bridging locally → REGRESSION; all-tunneled → F3 DRIFT / F4 n/a).

**Gate:** added the missing **EMPTY-SOURCE** verbatim prompt (yes/no, skips Stage 6.5).

**Contracts:** minimal Stage 9b path must build `stage1_*_records` via `_flatten` (so 2a–2e vars resolve + carry `_source_scope`); `gw_cluster_list` entries use the composite `<source_scope>/<cluster-name>` key matching `per_cluster`; **net_service is DEFERRED** so Stage 9/10 + the sample no longer emit `central_manage_net_services` (operator pre-populates); sample preview scope is a real Site, not Global; Act I-template cluster-mode refs are stage-neutral.

CI green; 1685 tests pass. I'll wait for your pass before calling it done.

nowireless4u:
Fixed in `2d05183` — and you're right, this one was my own bad fix from the previous round.

My W4 "fix" documented `per_vap.gw_cluster_list` as composite `<source_scope>/<cluster-name>` **key strings** and passed them straight to `central_translation_preview`. But the shipped `central:wlan_ssid` engine substitutes `gw_cluster_list` **whole** into the `overlay-wlan` body and expects the Central object shape `[{cluster, cluster-type, tunnel-type, cluster-scope-id}]` — which is exactly what it was built and live-tested with. Key strings would have emitted a structurally-wrong payload that *looks* valid.

The fix separates the two concerns:
- The decision stores **`cluster_refs`** — composite `per_cluster` keys, for collision-safe **lookup only**.
- The **2g snippet now resolves** each `cluster_ref` through `per_cluster` (to get the confirmed `central_scope_id`) and **builds the Central `gw_cluster_list` object list** before previewing. An unresolved/unconfirmed cluster skips with a `skip_reason` rather than emitting a bad binding.

Root cause worth noting: my holistic pass + the independent review I ran both checked the skill for *internal* consistency, but neither checked the doc against the **actual engine contract** (the `central:wlan_ssid` spec). Decision-model keys and engine-payload shapes are different things and must be converted at the boundary. I've added "verify skill pseudo-code against the real tool/engine contract, not just internal consistency" to my checklist.

CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — no blocking issues found.**

I reviewed the PR diff for `nowireless4u/hpe-networking-mcp#435`. This is a skill-markdown/docs/version bump change, and the new Stage 6.5 flow is generally consistent: source detection stays read-only, operator architecture decisions move after the verdict, provisional target-dependent findings no longer block the gate, and Stage 9b now guards missing decisions instead of indexing into empty maps. Solid cleanup.

### Critical
- None.

### Warnings
- None.

### Suggestions
- **`src/hpe_networking_mcp/skills/aos-migration.md:686`** — Stage 6.5 runs before Stage 7, but Q2 says to collect the concrete `central_scope_id` for each surviving cluster. Since Stage 7 is where target hierarchy mapping/confirmation happens, and some target scopes may not exist yet, consider clarifying that Stage 6.5 collects the intended Central scope **name/type** and later resolves it to `central_scope_id` or `<TBD:...>` during Stage 7/9b. As written, an agent may ask the operator for an ID that does not exist yet. Not blocking, but this is the one spot where the playbook could use sharper wording.

### Looks Good
- The split between verdict-gating findings and provisional target-dependent findings is clear and prevents the old dead-end where an operator could be blocked before choosing a different target architecture.
- Composite keys for VAPs and clusters (`<source_scope>/<name>`) are the right move. That avoids same-name collisions across AOS hierarchy scopes.
- Stage 9b’s new 2f/2g snippets handle missing Stage 6.5 decisions with explicit `skip_reason` output instead of throwing `KeyError` or silently defaulting to Global. That’s the kind of guardrail we like to see.
- The offline/upload path behavior is called out clearly: no `virtual_ap` / `ssid_prof` / `cluster_prof` coverage means skip Stage 6.5 and surface a translation gap.
- Version bump and docs/changelog updates are aligned with the new behavior.

### Verification run
- `uv run --frozen pytest tests/unit/test_skills.py tests/unit/test_skill_tool_references.py tests/unit/test_skill_aos8_live_detection.py`
  - **60 passed**
- `uv run --frozen pytest`
  - **1674 passed, 12 skipped**
  - No failures. The skipped count differs from the PR description, likely local environment/live-test availability, but the full collected suite passed clean here.

Nice work. No scuzzballs blocking the merge.

nowireless4u:
Thanks — closing the one non-blocking suggestion from your last pass in `3123ce8`.

You're right that Stage 6.5 (pre-Stage-7) can't ask for a concrete `scope_id` that may not exist yet. So:
- **Q2 now collects the intended Central scope NAME + type**, stored as `per_cluster[...]["target_scope_name"]` — not a numeric id.
- **2f and 2g resolve `target_scope_name` → `scope_id` via the walker with a `<TBD:...>` fallback** at preview time (the same resolution model the SSID scope in 2g already uses), and still skip with a `skip_reason` when the operator chose no scope at all — never a silent Global.

This also aligns the cluster path with the SSID path (both resolve a name → id/`<TBD>` in Stage 9b, since the target hierarchy is created during cutover). CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — one workflow gap to clean up before this gets relied on for dual-mode SSIDs.**

### ⚠️ Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md:671`, `:694-701`, `:1454-1464`** — Stage 6.5 runs before Stage 7 scope mapping/creation, but the `Bridged and Tunneled` path records and later requires `bridge_scope_id` / `tunnel_scope_id` directly.

  That’s a bad handoff. At Stage 6.5 time, the target Central scopes may not exist yet, and the rest of this PR correctly treats cluster placement as a **scope name** collected during the questionnaire and resolved later in Stage 9b with `<TBD:...>` fallback. Dual-mode SSID split scopes should follow the same pattern.

  **Suggestion:** collect `bridge_scope_name` / `tunnel_scope_name` in Stage 6.5, then resolve them in Stage 9b before calling `central_translation_preview`, e.g.:
  - `bridge_scope_id = scope_lookup.get(dec["bridge_scope_name"], f"<TBD:{dec['bridge_scope_name']}>")`
  - `tunnel_scope_id = scope_lookup.get(dec["tunnel_scope_name"], f"<TBD:{dec['tunnel_scope_name']}>")`

  Otherwise the dual-mode path can force the operator to provide IDs that may not exist yet, or it will skip preview with `bridged_and_tunneled needs both bridge_scope_id and tunnel_scope_id` even though the operator made a valid architecture choice by name.

### ✅ Looks Good

- The PR correctly separates **source readiness** from **target-architecture-dependent provisional findings**, which fixes the “BLOCKED before the operator can choose Bridged/controllerless” train wreck.
- Composite keys like `<source_scope>/<vap-name>` and `<source_scope>/<cluster-name>` are the right call. Same-name collisions across hierarchy scopes are real punks.
- Stage 9b now avoids silently defaulting cluster scope placement to Global. Solid.
- The new `gateway_cluster` and `wlan_ssid` preview wiring is clear and mostly defensive, especially the missing-decision and missing-SSID-profile `skip_reason` handling.
- Security posture is unchanged: markdown-only, no new write path, no credential handling changes.

### Checks run

- `gh pr diff 435 --repo nowireless4u/hpe-networking-mcp`
- `gh pr checks 435 --repo nowireless4u/hpe-networking-mcp` — all reported checks passing.
- Local subset:
  - `uv run --frozen python -m pytest tests/unit/test_skills.py tests/unit/test_skill_tool_references.py tests/unit/test_skill_snippet_sandbox_compat.py tests/unit/test_skill_aos8_live_detection.py`
  - Result: **75 passed**, 5 dependency warnings from `pyclearpass`.

Consider the dual-scope handoff the one scuzzball worth cleaning up.

nowireless4u:
```
rm -rf /tmp/hpe-networking-mcp-pr435 && gh repo clone nowireless4u/hpe-networking-mcp /tmp/hpe-networking-mcp-pr435 -- --quiet && cd /tmp/hpe-networking-mcp-pr435 && git fetch origin pull/435/head:pr-...
```
Reason: delete in root path

Reply `/approve` to execute, `/approve session` to approve this pattern for the session, `/approve always` to approve permanently, or `/deny` to cancel.

nowireless4u:
⏳ Working — 3 min — iteration 3/150, terminal

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested**

I reviewed the PR diff with `gh pr diff 435 --repo nowireless4u/hpe-networking-mcp`. This is skill/docs-only, no write-path code change, but there’s one correctness issue in the new Stage 9b WLAN preview wiring that should be cleaned up before merge.

### ⚠️ Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md` — Stage 9b / Step 2g WLAN SSIDs**
  
  The new WLAN preview snippet matches `virtual_ap` → `ssid_prof` by bare `profile-name`:

  ```python
  ssid_prof_names = {s.get("profile-name") for s in stage1_ssid_prof_records}
  sp_ref = (v.get("ssid_prof") or {}).get("profile-name")
  if sp_ref not in ssid_prof_names:
      ...
  ...
  "ssid_profiles": stage1_ssid_prof_records,
  ```

  That undercuts the cross-scope collision protection added elsewhere in this PR. Same-name `ssid_prof` objects can exist under different AOS 8 scopes with different settings. Passing the full flat list and relying on a bare-name join means `central_translation_preview` can select or validate against the wrong SSID profile when duplicate profile names exist across scopes.

  **Suggestion:** Make the playbook scope-aware here too. Either:
  - filter `ssid_profiles` to the effective profile set for the VAP’s `_source_scope`, or
  - key/resolve SSID profiles with a composite `<source_scope>/<ssid-profile-name>` identity before calling the preview engine, or
  - explicitly document that the engine must perform AOS hierarchy/effective-config resolution and reject ambiguous same-name profiles.

  Right now the surrounding text says `_source_scope` prevents collisions, but this join still lets one sneak through the side door. Some clown left the front door wide open.

### 💡 Suggestions

- **`src/hpe_networking_mcp/skills/aos-migration.md` — Stage 6.5 / Stage 9b terminology**
  
  The PR uses both operator-facing labels (`Bridged`, `Tunneled`, `Bridged-and-Tunneled`, `Hybrid`) and engine values (`bridged`, `tunneled`, `bridged_and_tunneled`, `hybrid`). Consider adding a small explicit mapping table in Stage 6.5 so future edits don’t accidentally pass display strings into `runtime_values["target_mode"]`.

### ✅ Looks Good

- The flow change is clear: detection recommends, operator decides.
- The BLOCKED verdict separation for target-independent vs provisional target-dependent findings is a solid fix.
- The new `skip_reason` guards are much better than indexing decisions blindly.
- Stage 9b now avoids silently defaulting unresolved scopes to Global, which is the right kind of paranoia.
- Changelog, docs table, and `pyproject.toml` version bump are aligned.

CI was still in progress when I checked the PR metadata, so this review is based on the diff, not completed checks.

nowireless4u:
Fixed in `322e51f` — and you've caught me at the same "apply it uniformly" miss again: my R9 fix made the *cluster* scope a name-resolved-later, but I left the dual-mode split scopes as raw ids. Aligned now:

- **Stage 6.5 collects `bridge_scope_name` / `tunnel_scope_name`** (Q1 dual row + the `per_vap` decision), not ids.
- **2g resolves both names → `bridge_scope_id` / `tunnel_scope_id`** via the walker with a `<TBD:...>` fallback before passing them to `central:wlan_ssid`; it guards on the *names* and skips with a `skip_reason` only when the operator set neither.

Scope resolution is now uniform across all three: cluster scope (`target_scope_name`), SSID scope (Stage-7 mapping), and the dual split scopes — every one resolves name → id/`<TBD>` in Stage 9b, since the target hierarchy is created during cutover.

CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — no blocking issues found.**

### Critical
- None.

### Warnings
- None.

### Suggestions
- **src/hpe_networking_mcp/skills/aos-migration.md:1029** — The translations table still says `central:wlan_ssid` uses `ssid_profiles` as the “full list,” but the new 2g implementation text correctly says to pass only the VAP’s own scope-filtered `ssid_prof` records. That table line should be updated to match the scope-aware join guidance so nobody follows the old bare-name/full-list pattern and reintroduces cross-scope collisions.

### Looks Good
- Stage 6.5 is placed after the readiness verdict and before Act II, which keeps the “source detects / operator decides” model clean.
- Composite `<source_scope>/<name>` keys for VAPs and clusters are the right guard against same-name objects across hierarchy scopes.
- The split between verdict-gating findings and provisional target-dependent findings is clearly documented, including the invariant that bucket-4 findings must not block the pre-questionnaire gate.
- Stage 9b’s new 2f/2g preview wiring does a good job avoiding silent Global fallbacks and surfacing missing decisions/scopes as `skip_reason`.
- EMPTY-SOURCE and offline/paste paths are handled explicitly instead of trying to ask architecture questions with missing inventory.

### Verification
- Ran `gh pr diff 435 --repo nowireless4u/hpe-networking-mcp`.
- Ran full local suite with `uv run --frozen pytest`: **1674 passed, 12 skipped**.

nowireless4u:
Fixed in `6e7b7ff`.

**Warning — scope-blind `ssid_prof` join.** You're right, that was the collision class sneaking in the JOIN. 2g passed the full flat `stage1_ssid_prof_records` to `central:wlan_ssid`, whose normalizer matches `virtual_ap.ssid_prof` → `ssid_prof` by bare `profile-name` — so a same-name profile at another scope (common under inheritance) could bind. Fixed by **passing only the VAP's own scope's `ssid_profiles`** (filtered by `_source_scope == v["_source_scope"]`); the missing-profile check is scoped too. The engine now joins within the correct scope, never across the full tree.

**Suggestion — label/value mapping.** Added an explicit note under the Q1 table: `Bridged`→`bridged`, `Tunneled`→`tunneled`, `Bridged and Tunneled`→`bridged_and_tunneled`, `Hybrid`→`hybrid` — store/pass the lowercase engine value into `runtime_values["target_mode"]`, never the display label.

CI green; 1685 tests pass.

For my own accounting: this is the same cross-scope-collision class you flagged for VAP/cluster *decision keys* — I'd applied composite-keying there but not to the `ssid_prof` *join*. I've now swept every place this PR passes a `stage1_*` list or matches by name into the engine; the scope-filtered SSID join is the last one (the cluster/VAP paths are composite-keyed, and the 2a–2e bare-record translations key by `record_id` per the existing design).

nowireless4u:
```
rm -rf /tmp/hpe-pr435-review && gh repo clone nowireless4u/hpe-networking-mcp /tmp/hpe-pr435-review -- --depth 1 >/dev/null && cd /tmp/hpe-pr435-review && gh pr checkout 435 && git diff --stat origin/...
```
Reason: delete in root path

Reply `/approve` to execute, `/approve session` to approve this pattern for the session, `/approve always` to approve permanently, or `/deny` to cancel.

nowireless4u:
Closed the lone R12 suggestion in `110d001`: the translations-catalog row described `central:wlan_ssid`'s `ssid_profiles` as the "full list" — contradicting the §2g scope-filtered join. Updated to "the VAP's own-scope records only — never the full flat list." CI green; 1685 tests pass.

nowireless4u:
⏳ Working — 3 min — iteration 4/150, terminal

nowireless4u:
⏳ Working — 6 min — iteration 9/150, receiving stream response

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — one docs/contract mismatch to clean up before this playbook is fully reliable.**

### ⚠️ Warnings

- **src/hpe_networking_mcp/skills/aos-migration.md:1450-1454** — The Stage 9b `central:wlan_ssid` snippet builds `gw_cluster_list` without `cluster-redundancy-type`, even though the same section documents the engine payload shape as including it (`[{cluster, cluster-type, tunnel-type, cluster-redundancy-type, cluster-scope-id}]`) and the shipped `central:wlan_ssid` translation target notes the same contract.  
  **Why it matters:** an operator following this snippet can generate overlay-wlan preview bodies with an incomplete gateway cluster binding. That punk may not show up in tests because the preprocessor currently passes `gw_cluster_list` through without validating the object shape.  
  **Suggestion:** either add the expected redundancy field to the constructed object, derived from the Stage 6.5/cluster decision, or explicitly update the documented engine contract if `cluster-redundancy-type` is intentionally optional.

### 💡 Suggestions

- The Stage 6.5 flow is much clearer than the old auto-derive wording. The composite `<source_scope>/<name>` decision keys and the explicit “don’t default missing scope to Global” guard are solid work.
- The EMPTY-SOURCE behavior and offline-parser skip path are spelled out well; that avoids a bunch of operator confusion.

### ✅ Looks Good

- Scope-aware SSID profile joins in 2g protect against same-name profile collisions across hierarchy scopes.
- Provisional target-dependent findings are separated from verdict-gating findings, so the operator does not get locked out of choosing a target architecture.
- GitHub checks are green: tests, lint/format, type check, Docker build, Bandit, dependency audit, and secret detection all passed.

nowireless4u:
Fixed in `369c276`. Good catch — `cluster-redundancy-type` is `x-mandatory: true` in the overlay schema (enum `PRIMARY`/`BACKUP`), and the 2g builder was omitting it.

Added it to the constructed object: the VAP's **first/anchor cluster = `PRIMARY`**, any additional (e.g. DMZ) clusters = `BACKUP` (`cluster_refs` is anchor-first; the operator confirms the primary/backup designation for multi-cluster designs).

Per the lesson from the last `gw_cluster_list` round, I **verified it live** rather than assuming: ran the `central:wlan_ssid` preview with the complete 5-field object and confirmed it round-trips into the `overlay-wlan` body as `[{cluster, cluster-type, tunnel-type, cluster-redundancy-type, cluster-scope-id}]`. CI green; 1685 tests pass.

nowireless4u:
```
rm -rf /tmp/hpe-networking-mcp-pr435 && gh repo clone nowireless4u/hpe-networking-mcp /tmp/hpe-networking-mcp-pr435 -- --filter=blob:none && cd /tmp/hpe-networking-mcp-pr435 && gh pr checkout 435
```
Reason: delete in root path

Reply `/approve` to execute, `/approve session` to approve this pattern for the session, `/approve always` to approve permanently, or `/deny` to cancel.

nowireless4u:
⏳ Working — 3 min — iteration 5/150, terminal

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — no blocking issues found.**

I reviewed the PR diff with:

```bash
gh pr diff 435 --repo nowireless4u/hpe-networking-mcp
```

### Critical
- None found.

### Warnings
- None found.

### Suggestions
- **CHANGELOG.md** — One line says Stage 6.5 Q2 records `central_scope_id`, but the new skill flow/schema records `target_scope_name` and resolves it later to `scope_id` in Stage 9b. The actual skill text is clear, so this is just wording cleanup to avoid confusing future readers.

### Looks Good
- The new Stage 6.5 flow is clearly separated from Act I readiness: detection recommends, operator decides.
- Good guardrail around target-dependent/provisional findings so they don’t incorrectly lock out the questionnaire.
- Composite keys using `<source_scope>/<vap-name>` and `<source_scope>/<cluster-name>` are the right move for cross-scope collision safety.
- Stage 9b preview wiring handles missing decisions/scopes with `skip_reason` instead of crashing or silently defaulting to Global. Now that’s solid work.
- The offline/minimal path is called out clearly, including skipping 2f/2g when required inventory isn’t available.

I did not post this review with `gh`; leaving that to the webhook delivery layer like the playbook says.

nowireless4u:
Closed the R14 suggestion in `9049805`: the CHANGELOG v3.3.7.0 entry said Stage 6.5 Q2 records `central_scope_id` per cluster — updated to the intended scope **name** (`target_scope_name`), resolved to `scope_id`/`<TBD>` in Stage 9b, matching the skill text. CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — one documentation-control-flow inconsistency should be fixed before merge. The feature direction looks solid, and the new Stage 6.5 flow is much clearer than the old auto-derive-and-apply framing.

### Critical
- None found.

### Warnings
- **src/hpe_networking_mcp/skills/aos-migration.md:1667** — The mandatory output rules still say verdicts are the “literal three values: **GO**, **BLOCKED**, or **PARTIAL**,” but this PR adds and documents **EMPTY-SOURCE** as a valid verdict in the gate logic and Act I template.
  - Why it matters: this is the “do not substitute alternatives” section, so an agent following the playbook may reject or reframe `EMPTY-SOURCE` even though the new flow depends on it.
  - Suggestion: update that bullet to include **EMPTY-SOURCE** as the fourth literal verdict.

### Suggestions
- **src/hpe_networking_mcp/skills/aos-migration.md:1627** — The decision matrix mentions “Stage 4a **A1–A3**” in the verdict-gating REGRESSION row, but A3 is defined as INFO-only earlier. Consider tightening that wording to “A1–A2” or “A1–A3 where severity is REGRESSION” so nobody reads AP-onboarding gap INFO as blocking.

### Looks Good
- Stage 6.5 is placed in the right spot: after readiness verdict, before Act II translation.
- Good separation between source-readiness findings and target-architecture-dependent provisional findings. That avoids the old dead-end where a target-mode regression could lock the operator out of choosing a different target.
- Composite keys for `<source_scope>/<vap-name>` and `<source_scope>/<cluster-name>` are the right play. Prevents cross-scope collisions.
- Stage 9b preview wiring is careful about scope resolution and avoids silently defaulting missing placement to Global. Now that’s solid work.
- Local verification passed: `uv run --frozen pytest tests -q` returned **1674 passed, 12 skipped**.
- GitHub checks are green: tests, lint/format, type check, Docker build, Bandit, dependency audit, and secret detection all passed.

nowireless4u:
Both fixed in `11af432`.

**Warning — EMPTY-SOURCE missing from the literal-verdicts rule.** The "do not substitute alternatives" output rule still said verdicts are the "three values GO/BLOCKED/PARTIAL" — omitting EMPTY-SOURCE, which the gate logic + Act I template rely on. Added it as the fourth literal value. (The other verdict lists already had all four; this was the one stale spot.)

**Suggestion — A1–A3 vs A1–A2.** The decision-matrix BLOCKED row said Stage 4a `A1–A3` gate the verdict, but A3 (AP-onboarding gap) is INFO-only. Corrected to `A1–A2`, and clarified in the Stage 4a bucket note that A3 is bucket-3/target-independent but **non-gating**.

CI green; 1685 tests pass.

nowireless4u:
```
rm -rf /tmp/hpe-networking-mcp-pr435-review && gh repo clone nowireless4u/hpe-networking-mcp /tmp/hpe-networking-mcp-pr435-review -- --depth=1
```
Reason: delete in root path

Reply `/approve` to execute, `/approve session` to approve this pattern for the session, `/approve always` to approve permanently, or `/deny` to cancel.

nowireless4u:
⏳ Working — 3 min — iteration 3/150, terminal

nowireless4u:
```
gh api repos/nowireless4u/hpe-networking-mcp/contents/src/hpe_networking_mcp/skills/aos-migration.md?ref=11af432c23528381cdcc0e2e57a44ccca4bf8381 --jq .content | python3 -c 'import sys,base64; print(b...
```
Reason: Security scan — [HIGH] Pipe to interpreter: gh | python3: Command pipes output from 'gh' directly to interpreter 'python3'. Downloaded content will be executed without inspection.

Reply `/approve` to execute, `/approve session` to approve this pattern for the session, `/approve always` to approve permanently, or `/deny` to cancel.

nowireless4u:
⏳ Working — 6 min — iteration 7/150, terminal

nowireless4u:
⏳ Working — 9 min — iteration 7/150, terminal

nowireless4u:
## Code review — Casey Jones

### Verdict: Approve

No blocking issues found in this PR.

### Reviewed scope

- `CHANGELOG.md`
- `docs/TOOLS.md`
- `pyproject.toml`
- `src/hpe_networking_mcp/skills/aos-migration.md`

### Checks

GitHub checks are green:

- Bandit Security Scan: pass
- Dependency Audit: pass
- Docker Build: pass
- Lint & Format: pass
- Secret Detection: pass
- Tests: pass
- Type Check: pass

### Findings

#### Critical
- None.

#### Warnings
- None.

#### Suggestions
- None blocking. The Stage 6.5 flow is a big markdown/runbook change, but the added guardrails around composite keys, provisional target-dependent findings, missing decisions, and avoiding silent Global scope defaults are solid work.

### Looks good

- Clear separation between **source-readiness verdict gating** and **target-architecture-dependent provisional findings**.
- Good correction of the old “auto-derived and applied” framing; the operator now explicitly decides the target architecture.
- Composite `<source_scope>/<object-name>` keys are called out for VAPs and clusters, which protects against same-name collisions across AOS scopes.
- Stage 9b preview wiring avoids silent bad defaults by emitting `skip_reason` when decisions or target scopes are missing.
- Empty-source behavior is explicitly documented.
- Version bump and docs/changelog updates are consistent with the skill-only change.

Consider yourself slam-dunked, scuzzball.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes recommended before merge** — the skill-only change is generally solid, and CI is green, but I found two runbook correctness issues that can cause the operator flow to collect decisions the preview cannot safely apply.

### ⚠️ Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md:688` — Stage 6.5 says to collect Central scope “name + type”, but the stored decision only keeps `target_scope_name`.**

  Some clown left a collision trap here. Central scope names are not safe as globally unique identifiers by themselves — a Site, Site-Collection, or Device Group can share a display name, and the same site name can appear under different parents. The text correctly says to confirm **name + type**, but the `decisions["per_cluster"]` schema only stores:

  ```python
  {"cluster_strategy": "...", "target_scope_name": "<scope-name>"}
  ```

  Then Stage 9b resolves with:

  ```python
  scope_lookup.get(dec["target_scope_name"], ...)
  ```

  That can resolve the wrong Central scope or become ambiguous.

  **Suggestion:** Store enough placement identity to disambiguate, for example:

  ```python
  {
    "cluster_strategy": "...",
    "target_scope_name": "...",
    "target_scope_type": "site|site_collection|device_group|global",
    "target_scope_path": "optional/full/path"
  }
  ```

  Then have Stage 9b resolve by path/type when available, not bare name only.

- **`src/hpe_networking_mcp/skills/aos-migration.md:665` — Q1’s “translatable VAP” definition is looser than the Stage 9b preview filter.**

  Q1 says a VAP is translatable if its `virtual_ap.ssid_prof` reference exists in the collected `ssid_prof` list. But the 2g preview loop later requires the SSID profile to exist in the **same `_source_scope`**:

  ```python
  scoped_ssid_profiles = [
      s for s in stage1_ssid_prof_records
      if s.get("_source_scope") == v.get("_source_scope")
  ]
  ```

  That mismatch can make Stage 6.5 ask the operator about a VAP that 2g will later skip because the matching `ssid_prof` only existed somewhere else in the hierarchy.

  **Suggestion:** Tighten the Q1 definition to match 2g exactly: the referenced `ssid_prof` must resolve in the VAP’s own `_source_scope`. That keeps the questionnaire and preview path lined up.

### ✅ Looks Good

- The new Stage 6.5 flow cleanly separates “source says what is” from “operator decides what they want.”
- Bucket-4 provisional findings are explicitly kept out of the pre-questionnaire verdict. That avoids locking out controllerless/bridged decisions before the operator gets a say.
- Composite keys for VAPs and clusters are the right move. Same-name objects across `/md` scopes are a real punk.
- Stage 9b’s new skip reasons are practical and safer than indexing missing decisions.
- CI/checks are green: tests, lint/format, type check, Bandit, dependency audit, secret detection, and Docker build all passed.

Consider yourself close, but I’d clean up those two scope-resolution/eligibility mismatches before calling this one done.

nowireless4u:
Casey Jones: PR #435 was closed, so no code review was needed.

PR #448, opened 2026-06-08
fix(skill): aos-migration scope identity + Q1/2g alignment (v3.3.7.1)

nowireless4u:
## Summary

Two follow-up fixes for warnings the automated reviewer raised on #435 **after it merged** (I merged before its review of the merge-of-main commit landed — my process miss). Both are real; skill-only, preview-only.

## 1. Central scope identity, not a bare name

`per_cluster` (and the dual-mode bridge/tunnel split scopes) recorded only `target_scope_name`, but a Central scope **name is not unique** — a Site, Site-Collection, and Device-Group can share a name, and the same name recurs under different parents. Name-only resolution can bind the wrong scope.

- Stage 6.5 now records the scope **name + type** (+ optional **path**): `target_scope_name` / `target_scope_type` / `target_scope_path` (and `bridge_scope_*` / `tunnel_scope_*` for dual mode).
- The Step-1 resolver `resolve(name, type, path)` (and `resolve_scope_id` wrapper) disambiguates: **path → name+type**, and returns **`<AMBIGUOUS:...>`** rather than silently picking the first of several matches.
- Applied **uniformly**: the per-cluster scope (2f), the dual bridge/tunnel scopes (2g), and the Stage-7-mapped SSID scope (2g) — all previously bare-name `scope_lookup.get`.

## 2. Q1 "translatable" now matches the 2g filter exactly

Q1 accepted a VAP whose `ssid_prof` reference existed **anywhere** in the collected list, but 2g requires it at the VAP's **own `_source_scope`** (the scope-filtered join). That mismatch could make Q1 ask the operator about a VAP that 2g then skips. Q1 now uses the same scoped match (`s._source_scope == vap._source_scope`).

## Testing
- Full suite: **1685 passed, 1 skipped**; ruff + format clean.
- Verified no residual bare-name `scope_lookup.get` in 2f/2g, and no new `central_*` tokens that would trip the skill-tool-reference test.

## Docs
- `CHANGELOG.md` 3.3.7.1; `pyproject.toml` 3.3.7.0 → 3.3.7.1.

Follow-up to #435.


nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — no blocking issues found.**

### ✅ Looks Good
- The PR fixes the bare-name scope lookup problem in the 2f/2g preview snippets by routing cluster, SSID, and dual-mode split scope resolution through `resolve_scope_id(name, type, path)`.
- The resolver now avoids silently picking the first same-name Central scope and returns `<AMBIGUOUS:...>` when name+type still isn’t enough. That’s the right failure mode — loud beats wrong.
- Q1’s “translatable VAP” definition now matches the 2g scoped `ssid_prof` join, so the questionnaire should no longer ask about VAPs that preview later skips.
- Version and changelog updates are consistent for `3.3.7.1`.
- I verified the changed skill snippets with targeted checks:
  - `tests/unit/test_skill_tool_references.py`
  - `tests/unit/test_skill_snippet_sandbox_compat.py`
  - Result: `30 passed`

### 💡 Suggestions
- Consider making the Stage 7 mapping contract explicit in the skill output table or nearby text. The new 2g snippet assumes `stage7_central_scope_for()` returns an identity object like `{"name": ..., "type": ..., "path": ...}`, but the Stage 7 table still mainly documents `Target type` and `Target name`. Adding a short note that Stage 7 must retain/derive full Central path from the confirmed hierarchy would keep future readers from rebuilding a bare-name lookup by accident.

### Notes
- No security concerns found in this docs/skill-only change.
- No residual `scope_lookup.get` calls remain in the reviewed 2f/2g area.

nowireless4u:
Closed the lone suggestion in `7829c8e`: added an explicit **Stage 7 mapping contract** note — each mapping row retains the scope **name + type + full Central path** (not just a name), and `stage7_central_scope_for()` returns that `{name, type, path}` identity so resolution is unambiguous. Calls out explicitly not to collapse a row to a bare name (which would rebuild the cross-scope-collision bug). CI green; 1685 tests pass.

nowireless4u:
## Code review — Casey Jones

**Verdict: Changes requested** — one correctness issue in the new scope identity resolver. The idea is solid, but one punk slipped through the line.

### ⚠️ Warnings

- **`src/hpe_networking_mcp/skills/aos-migration.md:688`, `:713`, `:1087-1088`** — `target_scope_type` documents `device_group`, but `central_get_scope_tree` returns device-group scopes as `DEVICE_COLLECTION`.

  The resolver lowercases and compares directly:

  ```python
  (n.get("type") or "").lower() == scope_type.lower()
  ```

  So an operator-provided `target_scope_type="device_group"` will not match a Central tree node with `type="DEVICE_COLLECTION"`. That means valid Device Group placements can resolve to `<TBD:...>` instead of the actual `scope_id`, undercutting the exact scope-disambiguation fix this PR is trying to make.

  **Suggestion:** normalize user-facing Stage 6.5 type values to Central tree type tokens before filtering, for example:

  ```python
  TYPE_ALIASES = {
      "global": "global",
      "site": "site",
      "site_collection": "site_collection",
      "site-collection": "site_collection",
      "device_group": "device_collection",
      "device-group": "device_collection",
      "device_collection": "device_collection",
  }
  wanted = TYPE_ALIASES.get(scope_type.lower(), scope_type.lower())
  m = [n for n in m if (n.get("type") or "").lower() == wanted]
  ```

  Or update the documented stored value from `device_group` to `device_collection` everywhere. I’d prefer aliasing so the operator-facing language can stay “Device Group” without breaking the engine.

### 💡 Suggestions

- **`src/hpe_networking_mcp/skills/aos-migration.md:1080-1084`** — If `scope_path` is provided but does not match exactly, the resolver immediately returns `<TBD:scope_path>` and never falls back to `name + type`. That may be intentional, but the surrounding text says “path → name+type,” which reads like a priority order. If exact-path mismatch should be terminal, clarify that in the playbook. If not, fall through to the name/type logic after the path miss.

### ✅ Looks Good

- Q1 now matches the 2g scoped SSID-profile join instead of accepting same-name profiles from another scope. That closes the questionnaire/preview mismatch.
- The remaining bare `scope_lookup.get(...)` uses in the changed 2f/2g flow are gone.
- The per-cluster, dual bridge/tunnel, and Stage-7 SSID scope paths now consistently carry identity instead of just a bare name.
- Version bump and changelog entry are aligned.

### Checks run

- `gh pr diff 448 --repo nowireless4u/hpe-networking-mcp`
- `uv run --frozen pytest tests/unit/test_skill_tool_references.py tests/unit/test_skill_snippet_sandbox_compat.py`

Result: **30 passed**, with unrelated dependency `SyntaxWarning`s from `pyclearpass`.

nowireless4u:
```
gh api repos/nowireless4u/hpe-networking-mcp/contents/src/hpe_networking_mcp/skills/aos-migration.md?ref=6a2c1772867f0b1d73764f1ae907f620ff01e450 --jq .content | base64 -d | python3 - <<'PY'
import sy...
```
Reason: Security scan — [HIGH] Pipe to interpreter: base64 | python3: Command pipes output from 'base64' directly to interpreter 'python3'. Downloaded content will be executed without inspection.; [HIGH] Base64 decode piped to interpreter: Command decodes base64 content and pipes it directly to an interpreter for execution

Reply `/approve` to execute, `/approve session` to approve this pattern for the session, `/approve always` to approve permanently, or `/deny` to cancel.

nowireless4u:
⏳ Working — 3 min — iteration 3/150, terminal

nowireless4u:
## Code review — Casey Jones

**Verdict: Comment — no blocking code-review findings found in the diff.**

### Critical
- None found.

### Warnings
- None found.

### Suggestions
- **CI is still pending** at review time for tests, lint/format, type check, Bandit, dependency audit, and Docker build. Secret detection has passed. I’d wait for the remaining checks before merging.

### Looks Good
- The scope identity fix addresses the real collision risk: scope name alone is not safe when Central can have matching names across scope types or parent paths.
- `resolve_scope_id(name, type, path)` now prefers path, falls back to name+type, and returns `<AMBIGUOUS:...>` instead of silently picking the first match. That’s the right behavior for preview tooling.
- The 2f gateway cluster path, 2g SSID scope mapping, cluster overlay binding, and bridged/tunneled split scopes now consistently use identity-based resolution instead of bare `scope_lookup.get(...)`.
- Q1’s “translatable VAP” definition now matches the 2g scoped join, which fixes the operator-questionnaire mismatch.
- Changelog and version bump are aligned with the patch scope.

No scuzzballs spotted in the diff. Just let CI finish its patrol.

PR #462, opened 2026-06-10
feat(skill): central-ucc-quality — UCC call-quality correlation + Generative-UI dashboard (closes #400)

### LINKED ISSUE #400: Feature request: UCC/call quality tool via show ucm cdrs
Summary: Add a get_ucc_status tool (or guided prompt) that runs show ucm cdrs across APs and returns parsed UCC call quality data.
Background: Tested today with Claude against a Redondo site running AOS-10 AP-735s. central_show_commands with show ucm cdrs successfully returns live CDR tables including per-stream delay, jitter, packet loss, and UCC scores for WiFi Calling (VoWiFi), Zoom, and MS-Teams-Voice sessions. The command works on AOS-10 APs, not just gateways.
Suggested tool behavior:

Accept a list of AP serials or a site name
Fan out show ucm cdrs concurrently across APs
Parse and return active streams with quality metrics
Flag streams exceeding ITU G.114 thresholds (delay >150ms, jitter >30ms, loss >1%)

Why it matters: UCC visibility is a top use case for wireless troubleshooting. This data is already in the AP — it just needs a tool wrapper. The existing central_show_commands tool makes this possible today, but a dedicated tool with parsing and threshold logic would make it much more useful for AI-driven diagnosis.
Also worth considering: Dynamic tool discovery / skills/runbook support so complex multi-AP workflows like this don't require hardcoded tools for every use case.

nowireless4u:
## Summary

New bundled skill **`central-ucc-quality`** — trustworthy UCC (voice/video) call-quality on Aruba Central AOS-10 APs and gateways, with an optional Generative-UI dashboard. **Closes #400.**

The skill correlates three live UCM tables from a **single atomic `central_show_commands` snapshot** (so all three describe the same instant):

| Command | Role |
|---|---|
| `show ucm hashtable` | client **IP → MAC** join (identity only) |
| `show datapath session ucc` | **liveness source of truth** — a stream is live only if its IP 5-tuple appears here with a populated `Codec:` field |
| `show ucm cdrs` | the metrics — delay / jitter / loss / **UCC Score** |

### Why a skill and not a hardcoded tool
Per the #400 design discussion + live AOS-10 capture: the CDR is a *Call Detail Record log that persists after calls end*, with **no timestamp and no active/ended flag** — so a torn-down call is byte-for-byte indistinguishable from a live degraded one. A tool wrapping `show ucm cdrs` alone reports **phantom "active calls" with stale metrics**. The value is the **join + stale-vs-live discrimination** (datapath `Codec:` = liveness; `Callinfo list:N > 0` does *not* prove liveness — a client was observed frozen at `Callinfo:2` with dead sessions). That's a "correlate three commands" problem — a good fit for a runbook, not a single tool.

### Design rules encoded (from live capture)
- **Liveness** decided by `show datapath session ucc` (populated `Codec:`), never by the CDR. Non-live CDR rows are surfaced separately as stale records.
- **Join key** is the IP 5-tuple; CDR `src_ip` = client IP → MAC via the hashtable.
- **Quality gated on the UCC Score** (≥80 good / 70–79 fair / <70 poor), deliberately **not** on the Delay field — Delay reads thousands of ms (≈9000 ms observed), is not one-way latency, and ITU-T G.114's 150 ms threshold would flag nearly every record. Delay/jitter/loss shown as context.
- One atomic comma-separated snapshot per device. **Read-only.**

### Dashboard
Renders an interactive Prefab dashboard via `generate_prefab_ui` (KPI cards + worst-first data table with the UCC Score column color-coded + per-app/per-band charts) when the server runs with `MCP_APP_ENABLE=true`. Falls back to an emoji-coded Markdown table otherwise — never errors just because the app tool is absent. The skill spells out the **data contract** (values in `data={...}` become Prefab globals *by key*, not a `data` variable) to avoid the `NameError` trap.

## Scope
- AOS-10 APs (`device_type="aps"`) primary; gateways supported.
- AOS-8 controllers out of scope (different UCM CLI / `aos8_*` tools).
- Runbook-only: **no new tools, no code paths altered.**

## Testing
- Both skill-hygiene regression tests pass: sandbox-snippet compatibility (no `re`/`collections`/`async def`/`yield`/OS-access in any `python` block) and tool-reference resolution.
- Full suite green in Docker: **1751 passed, 1 skipped**; ruff + format + mypy clean.

## Docs
- `docs/TOOLS.md` — Bundled skills table row.
- `INSTRUCTIONS.md` — skill trigger-phrase table row.
- `CHANGELOG.md` — 3.3.10.0 entry.
- `pyproject.toml` — version 3.3.10.0 (minor: new capability).


nowireless4u:
Thanks Casey — all three addressed in b54e57a.

**Blocking — 4-tuple vs 5-tuple:** you're right, and the honest fix is to stop overclaiming. `show ucm cdrs` emits no protocol column, so the CDR↔datapath correlation can only be an **IP/port 4-tuple** match. I reworded rule 1, rule 3, and the command table to say 4-tuple, and added the justification: UCC media is UDP (RTP/SRTP, IKE/4500) so the 4-tuple is effectively unique, the residual collision (a stale CDR row sharing an exact `src:port→dst:port` with a *live different-protocol* flow) isn't a real-world condition, and `show datapath session ucc` *does* carry protocol (col 3, `17`=UDP) so anyone wanting a strict match can filter the live set to `proto == "17"`. The parser now keeps `proto` bound for exactly that.

**Warning — undefined `blocks_for`:** replaced with a defined, sandbox-safe `pick_blocks(resp)` that normalizes the `central_show_commands` response across the shapes it returns (list of `{command, output}` / dict-keyed-by-command / single concatenated string) into the three raw outputs. If it can't split per-command it returns the whole blob, which is safe here because every parser is line-filtered (mac/IP lines, `Codec:` lines, metric rows) — a concatenated output still parses correctly. Documented that.

**Suggestion — dashboard availability:** added an explicit note at Step 4 that the check completes at Step 3 and `generate_prefab_ui` absence only changes *rendering* (Markdown fallback), never marks the UCC check failed.

Both skill-hygiene regression tests still pass (sandbox-snippet compat + tool-reference resolution).

Issue #444, opened 2026-06-07
GreenLake tools crash instead of returning 503 when GreenLake is not configured


nowireless4u:
## Bug description
GreenLake tools assume `config.greenlake` and `greenlake_token_manager` are present. If GreenLake is disabled or startup failed, they can crash with `AttributeError` instead of returning a clear 503-style ToolError.

## Evidence
- `src/hpe_networking_mcp/server.py:97-101` sets `context["greenlake_token_manager"] = None` when GreenLake initialization fails or is not configured.
- `src/hpe_networking_mcp/platforms/greenlake/tools/users.py:108-113` reads `greenlake_token_manager`, dereferences `config.greenlake.api_base_url`, constructs `GreenLakeHttpClient`, then calls the API.
- `src/hpe_networking_mcp/platforms/greenlake/client.py:91-95` assumes `token_manager` has `get_auth_headers()`.

## Why it matters
A disabled or failed GreenLake integration should produce an actionable "GreenLake is not configured" error. Instead, current paths can raise internal attribute errors, which are harder for users and agents to recover from.

## Suggested fix
Add a shared `get_greenlake_client(ctx)` / `get_greenlake_context(ctx)` helper that checks both `config.greenlake` and `greenlake_token_manager` before constructing clients.

Return a `ToolError` with `status_code: 503` and a clear credential/configuration message, then apply it across GreenLake tools.

PR #446, opened 2026-06-08
ci: add Dependabot and enforce dependency audit

nowireless4u:
## Summary
- Add Dependabot config for weekly pip/uv dependency update PRs.
- Update the existing Security workflow so pip-audit runs against the frozen lockfile and fails on unignored vulnerabilities.
- Refresh uv.lock to fix the advisories that can be fixed with current dependency constraints.

## Safety model
Dependabot opens PRs only. No auto-fix and no auto-merge are enabled. Dependency updates still need review and test validation before merge.

## Remaining dependency note
- requests==2.32.5 is still pinned by pycentral==2.0a17, so requests>=2.33.0 is currently unsatisfiable.
- The workflow explicitly ignores CVE-2026-25645 with an inline comment because the advisory only affects direct use of requests.utils.extract_zipped_paths(). Remove the ignore once pycentral allows requests>=2.33.0.

## Test plan
- [x] uv run --frozen pip-audit --ignore-vuln CVE-2026-25645
- [x] uv run --frozen ruff check .
- [x] uv run --frozen mypy src
- [x] uv run --frozen bandit -q -r src -c pyproject.toml
- [x] uv run --frozen pytest -q — 1674 passed, 12 skipped

Refs #445

Issue #445, opened 2026-06-08
Dependency vulnerabilities found by pip-audit


nowireless4u:
## Summary

`uv run --frozen pip-audit` reports 11 known vulnerabilities across 6 packages in the current locked environment.

This is from the committed `uv.lock` environment, not from an unfrozen dependency refresh.

## Command run

```bash
uv run --frozen pip-audit
```

## Findings

| Package | Locked version | Advisory | Fixed version |
| --- | ---: | --- | ---: |
| idna | 3.13 | CVE-2026-45409 / GHSA-65pc-fj4g-8rjx | 3.15 |
| pip | 26.1 | PYSEC-2026-196 / CVE-2026-8643 | 26.1.2 |
| pyjwt | 2.12.1 | PYSEC-2026-179 / CVE-2026-48526 / GHSA-xgmm-8j9v-c9wx | 2.13.0 |
| pyjwt | 2.12.1 | PYSEC-2026-175 / CVE-2026-48522 / GHSA-993g-76c3-p5m4 | 2.13.0 |
| pyjwt | 2.12.1 | PYSEC-2026-177 / CVE-2026-48524 / GHSA-fhv5-28vv-h8m8 | 2.13.0 |
| pyjwt | 2.12.1 | PYSEC-2026-178 / CVE-2026-48525 / GHSA-w7vc-732c-9m39 | 2.13.0 |
| requests | 2.32.5 | CVE-2026-25645 / GHSA-gc5v-m9x4-r6x2 | 2.33.0 |
| starlette | 1.0.0 | PYSEC-2026-161 / CVE-2026-48710 / GHSA-86qp-5c8j-p5mr | 1.0.1 |
| urllib3 | 2.6.3 | PYSEC-2026-142 / CVE-2026-44432 / GHSA-mf9v-mfxr-j63j | 2.7.0 |
| urllib3 | 2.6.3 | PYSEC-2026-141 / CVE-2026-44431 / GHSA-qccp-gfcp-xxvc | 2.7.0 |

Note: `pip-audit` prints the Starlette advisory twice for the same package/version, so the CLI summary says 11 vulnerabilities. The unique Starlette advisory is listed once above.

## Dependency paths observed in `uv.lock`

- `idna`: pulled by `anyio`, `email-validator`, `httpx`, `requests`
- `pip`: pulled by `pip-api`
- `pyjwt`: pulled by `mcp[crypto]`
[...]

Issue #443, opened 2026-06-07
Central tools raise AttributeError instead of 503 when Central client is unavailable


nowireless4u:
## Bug description
When Central is not configured or failed initialization, Central tools can raise internal `AttributeError` instead of returning a clear unavailable/configuration error.

## Evidence
- `src/hpe_networking_mcp/server.py:80-84` sets `context["central_conn"] = None` when Central initialization fails or Central is not configured.
- Central tools pass `ctx.lifespan_context["central_conn"]` directly to helpers; for example `src/hpe_networking_mcp/platforms/central/tools/alert_configs.py:180-186`.
- `src/hpe_networking_mcp/platforms/central/utils.py:308-313` calls `central_conn.command(...)`.
- `src/hpe_networking_mcp/platforms/central/utils.py:314-323` then calls `central_conn.logger.error(...)` inside the exception handler.

## Why it matters
If `central_conn` is `None`, the original failure is `None.command`, then the exception handler can also fail on `None.logger`. Users get an internal tool/runtime error instead of a clear "Central is not configured" response.

## Suggested fix
Add a shared `get_central_conn(ctx)` helper that raises `ToolError({"status_code": 503, ...})` when Central is unavailable, and use it in Central tools.

Also harden `retry_central_command` so its error handler does not assume `central_conn.logger` exists.

nowireless4u:
Fixed in #463 (merged to main). Released in v3.3.10.1.

PR #463, opened 2026-06-10
fix: harden unavailable-platform paths + async GreenLake token + PII show-output sweep (closes #444, #443, #442, #411, #440)

nowireless4u:
## Summary

A batch of fixes from the `bug`-label triage. All are defensive hardening or isolated correctness fixes — **no new tools, and no behavior change for correctly-configured platforms.**

| Issue | Fix |
|---|---|
| **#443** | Central tools raised an opaque `AttributeError` when Central was unavailable |
| **#444** | GreenLake tools crashed instead of returning 503 when not configured |
| **#440** | GreenLake OAuth token acquisition blocked the async event loop |
| **#442** | Mist `getSelf` startup resolver crashed on an HTTP 200 non-JSON body |
| **#411** | PII tokenization skipped raw show-command CLI output |

## What changed

**#443 — Central unavailable → clear 503, not `AttributeError`**
- New shared `get_central_conn(ctx)` helper raises `ToolError({"status_code": 503, ...})` when `central_conn` is `None` (not configured / failed startup).
- **Every** Central tool's connection access now routes through it (the previous direct `ctx.lifespan_context["central_conn"]` dereferences crashed on `None`).
- `retry_central_command` raises the same 503 up front on a `None` connection, and logs transport errors via a safe logger that no longer dereferences `central_conn.logger` inside the `except` block — the original raised a **second** `AttributeError` there.

**#444 — GreenLake unavailable → clear 503**
- New `get_greenlake_client(ctx)` validates both `config.greenlake` and the token manager before building a client; raises 503 otherwise. Applied across all GreenLake tools.

**#440 — GreenLake token off the event loop**
- The synchronous OAuth fetch/refresh now runs via `asyncio.to_thread` on the request path, and the startup `TokenManager` (which does a blocking initial fetch in `__init__`) is constructed off-loop in the lifespan handler. (GreenLake's actual API transport was already `httpx.AsyncClient` — only the token was sync.)

**#442 — Mist `getSelf` tolerates HTTP 200 non-JSON**
- A proxy/WAF/login-page interstitial returning `200` with HTML no longer raises during org-id resolution: the JSON decode is guarded, a bounded body preview is logged, and the resolver fails closed to `None`.

**#411 — PII sweep over show-command output**
- The `output` blob from show-command tools (`central_show_commands` etc.) classified `SKIP`, so only the universal email/AWS-URL scan ran over it. It's now a free-text field, so the full `_scan_free_text` sweep (PEM cert/key blocks, emails, MAC normalization) runs over CLI dumps when `ENABLE_PII_TOKENIZATION=true`.
- **Honest scope note:** field-name-keyed secrets (PSKs, shared secrets) have no value-shape regex and are not detectable inside an opaque CLI blob — that needs a dedicated CLI-grammar sweep and is out of scope here. Documented inline.

## Testing
- New `tests/unit/test_unavailable_platform_guards.py` covers the Central/GreenLake 503 paths, the safe transport logger, the GreenLake async header builder, and the Mist non-JSON resolver.
- New PII cases in `test_pii_redaction.py` assert PEM blocks / emails / MACs in a `show`-output blob are swept end-to-end.
- Full suite green in Docker: **1769 passed, 1 skipped**; ruff + format + mypy clean.

## Deliberately not in this PR
- **#43** (VoIP port-bounce edge-port recognition) — excluded at maintainer's request.
- **#441** (ClearPass blocking async) — the entire pyclearpass transport is synchronous across ~384 call sites (vs GreenLake, where only the token was sync). De-blocking it is a transport-wide refactor that warrants its own focused PR, not this batch.
- **#445** (dependency CVEs) — the dependabot + audit infra already merged in #446; the remaining `uv.lock` refresh is a separate change.
- The confirmation-gating trio (#415/#416/#436) is a security rework tracked separately.


nowireless4u:
Thanks Casey. Picked up your GreenLake token-threading suggestion in ba102ce — added double-checked locking around `TokenManager._ensure_valid_token` (a `threading.Lock` seeded in `__init__`) so concurrent off-loop `get_auth_headers` calls don't stampede the token endpoint when the cached token expires; only one thread refreshes and the rest observe the freshly-set token. Added a barrier-synchronized regression test (8 concurrent callers → exactly one fetch).

The PII `output`-as-free-text note and the out-of-scope CLI-secret caveat: agreed, kept those explicit in the code + changelog. CI was green on the prior push; this follow-up re-runs it.

Issue #466, opened 2026-06-11
Platform structure conformance: standardize all platforms on the _template shape


nowireless4u:
## Goal

Every platform should be built from the same structural template so maintenance work (async/threading, token management, elicitation/capability gating, error contract) is one pattern applied N times — not N bespoke designs. `platforms/_template/` is the canonical shape:

- `client.py`: `<Platform>Client` class wrapping `httpx.AsyncClient` + `<Platform>AuthError` + async `get_<platform>_client()` accessor + `format_http_error()`; ToolError contract.
- `_registry.py`: `tool = make_tool_decorator("<platform>", lambda: mcp)` (shared factory, no local shim).
- `tools/`: `@tool(capability=...)` per the annotation rubric; no hand-written `ToolAnnotations`.

## Conformance audit (2026-06-11, current main)

| Platform | Client shape | Registry | Annotations | Verdict |
|---|---|---|---|---|
| apstra | `ApstraClient` + accessor + `format_http_error` | local shim | hand `ToolAnnotations`/tags | client ✅ / registry+annotations pending |
| axis | `AxisClient` + accessor + `format_http_error` | shared factory | `capability=` | ✅ fully conformant |
| uxi | `UXIClient` + accessor + `format_http_error` | local shim | hand `ToolAnnotations`/tags | client ✅ / registry+annotations pending |
| aos8 | `AOS8Client` but **no `get_aos8_client()` accessor, no `format_http_error`, own `AOS8APIError` instead of ToolError in client layer** | shared factory | `capability=` | client deviates |
| greenlake | `GreenLakeHttpClient` but **accessor takes `ctx` arg (others use `get_context()` internally); sync `TokenManager` in separate `auth.py`, wrapped in `asyncio.to_thread`** | local shim | dict-style hints (no `ToolAnnotations` class) | client deviates |
| mist | **functional style: `build_mist_client()` + module-level `mist_request()`, no client class, `_client.py` naming** (generated platform) | local shim | 1037 generator-emitted `ToolAnnotations` | deviates (decide: conform vs documented exception) |
| central | pycentral SDK (sync, no httpx) | local shim | mixed (`ToolAnnotations` + module constants + tags) | rebuild → #464 |
| clearpass | pyclearpass SDK (sync, monkey-patched) | local shim | tags + inline confirms | rebuild → #465 |

[...]

Issue #465, opened 2026-06-11
ClearPass: replace pyclearpass with a template-conformant async httpx client (bundle with capability migration)


nowireless4u:
## Motivation

The pyclearpass SDK costs more than it provides:

1. **We already maintain its auth layer ourselves.** `platforms/clearpass/client.py` monkey-patches `ClearPassAPILogin._send_request` to add token refresh-on-401/403 (the SDK has none — tokens expire after 8h) and to honor `verify_ssl` (the SDK hardcodes `False`). Any pyclearpass release can silently invalidate that patch.
2. **All ClearPass calls are synchronous and block the event loop** — reported in #441. Both the OAuth refresh (`httpx.post`) and every patched `_send_request` call run sync on async tool paths.
3. **What remains of the SDK is URL templates.** Each pyclearpass API method builds a path and calls `_send_request`. With the vendored API specs as source of truth, that's knowledge we already hold.

## Proposed change

Rebuild on `httpx.AsyncClient` following `platforms/_template/client.py` (same shape as Apstra/Axis/UXI):

- `ClearPassClient` class + module accessor + `format_http_error()`, async token manager with refresh-on-auth-failure (replacing both `ClearPassTokenManager` and the `_send_request` monkey-patch), configurable TLS verification.
- Rewrite tool call sites: ~38 tool modules currently instantiate per-area pyclearpass API classes (`ApiPolicyElements`, `ApiIdentities`, `ApiGlobalServerConfiguration`, …). Each call reduces to method + path + body against the new client.
- Remove `pyclearpass` from `pyproject.toml`.
- Regression test: a slow ClearPass request must not block concurrent async tool work (carried over from #441).

## Bundle with the capability-annotation migration

The capability migration (#415 / #416 groundwork, rubric from the annotations factory PRs) must also touch every ClearPass tool file. Do both in one per-file sweep so each of the ~38 files is touched once:

1. Replace the pyclearpass call with the new async client call.
2. Migrate the decorator to `capability=` per `docs/tool-annotation-rubric.md`; switch `_registry.py` to the shared `make_tool_decorator` factory.
3. Delete the inline confirmation machinery (~80 inline confirm defs/calls) that the universal `requires_confirmation` gate will subsume.

## Relationship to other work

- Subsumes #441 (its "replace the ClearPass transport with async HTTP" option is this issue).
- Advances the per-platform migration sequence toward #415 / #416.
[...]

Issue #464, opened 2026-06-11
Central: replace pycentral with a template-conformant async httpx client


nowireless4u:
## Motivation

The pycentral SDK no longer pays its way and actively costs us:

1. **Every Central call blocks the event loop.** All ~660 Central tools funnel through `retry_central_command()` (`platforms/central/utils.py`), which calls the synchronous `central_conn.command(...)` directly from async tool functions — no thread offload on the main tool path. Only the cross-platform statics (`site_health_check`, `site_rf_check`) wrap pycentral in `asyncio.to_thread`. A slow Central API call stalls the entire server, the same defect class fixed for GreenLake token fetch in #463 and reported for ClearPass in #441.
2. **Alpha dependency pin.** `pycentral==2.0a17` — an alpha pin in a production dependency tree, right after Dependabot enforcement landed (#446).
3. **The SDK surface we use has shrunk to almost nothing.** Three direct `conn.command()` call sites plus a handful of `pycentral.new_monitoring` helper classes (`MonitoringSites`, `MonitoringAPs`, `MonitoringGateways`, `Clients`, `Troubleshooting`). Everything else already goes through our own chokepoint, our own retry logic, our own pagination, and our own error contract. The vendored OpenAPI specs are now the source of truth for the tool surface (#423), so the SDK's endpoint knowledge is redundant.
4. **Blocks upcoming platform work.** The capability-annotation migration (#415/#416), elicitation gating, and centralized token management all touch this platform; carrying a sync SDK through that work means doing it twice.

## Proposed change

Rebuild the Central transport on `httpx.AsyncClient` following the `platforms/_template/client.py` shape (same pattern as Apstra/Axis/UXI):

- `CentralClient` class + module accessor + `format_http_error()`, ToolError error contract.
- Async OAuth2 client-credentials token manager (no `asyncio.to_thread` workaround).
- Swap the internals of `retry_central_command()` — this converts ~660 tools in one move since they all flow through it.
- Replace the 3 direct `conn.command()` call sites and the `new_monitoring` helper usages (`site_health_check.py`, `site_rf_check.py`, `scope_builder.py`, and a few tool modules) with plain path calls; drop the now-unneeded `asyncio.to_thread` wrappers.
- Remove `pycentral` from `pyproject.toml`.
- Tests: reuse the established httpx `_transport`-swap mock pattern; add a regression test that a slow Central request does not block concurrent async tool work.

## Verification caveat

[...]

PR #467, opened 2026-06-11
refactor(auth): shared AsyncTokenManager primitive; UXI + Apstra + template adopt (v3.3.10.2)

nowireless4u:
## What

First step of the platform-standardization effort (References #466 — step 1 of its proposed sequence; intentionally **not** a closing keyword, the effort is multi-PR).

Extracts the two proven in-tree token lifecycles into a shared primitive, `platforms/_common/auth.py`:

- **`AsyncTokenManager`** — cached token + expiry buffer + `asyncio.Lock` with double-checked locking; `get_token()` / `refresh()` / `invalidate()` / `prime()` / `static()`.
- **`TokenResult`** — one acquisition's outcome; `expires_in=None` models login-session tokens that live until a 401.
- **`AuthError`** — shared base; platform flavors subclass it (`ApstraAuthError`, `TemplateAuthError`).
- **`oauth2_client_credentials(...)`** — fetch-strategy factory for the client-credentials grant (extracted from UXI, including the fresh no-base_url client for the token POST).

The split of responsibilities: the manager owns *when* to acquire/cache/invalidate; the platform supplies *what request* acquires the token (fetch strategy) and *how the credential attaches* to requests (header/cookie/query stays in the client — this is what lets AOS 8's UIDARUBA flavor adopt later without contorting the primitive).

## Adopters in this PR

- **UXI** (OAuth2 donor): drops inline `_ensure_token` / `_fetch_token_locked` / `_invalidate_token`; wires `oauth2_client_credentials` into a manager with the existing 60 s buffer and 7199 s default TTL.
- **Apstra** (login-session donor): `_login()` becomes the manager's fetch strategy returning `TokenResult(token)`; 401 → `refresh()` retry preserved.
- **`_template/client.py`**: rewritten as the canonical consumer (static-token strategy by default, docstrings point to the OAuth2/login-session flavors).

Public client surfaces are unchanged: `request()`, `get_json()`, `uxi_get/post/patch/delete`, `health_check()`, accessors, `format_http_error` all keep their signatures and behavior.

## Why extraction, not new design

Both donors already implement exactly the lifecycle every platform needs (UXI: time-based expiry + lock-serialized refresh; Apstra: re-login-on-401 under lock). The primitive is their union, with no behavior changes — so the existing client test suites double as regression coverage for the extraction.

## Follow-ups (per #466 sequencing)

1. GreenLake adopts (its 290-line sync OAuth2 provider collapses into the shared strategy; removes the #463 `to_thread` band-aid).
2. AOS 8 adopts the login-session flavor (cookie + query-param attachment stays in its client).
3. Central (#464) and ClearPass (#465) rebuilds target the primitive directly — no new per-platform token managers.

## Tests

- New `tests/unit/test_common_auth.py` (16 tests): lifecycle (cache / buffer / no-expiry / invalidate / refresh / concurrent single-fetch / empty-token guard), OAuth2 fetcher (form encoding, default expiry, non-2xx propagation, missing `access_token`), and manager+fetcher end-to-end.
- UXI/Apstra unit tests updated to target the manager API (they previously monkeypatched the now-removed private internals).
- Full suite in Docker: **1786 passed, 1 skipped**; ruff + format + mypy + bandit clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.3.10.2] - 2026-06-11
+
+**Patch — shared async auth primitive (`platforms/_common/auth.py`); UXI + Apstra + `_template` adopt it.** First step of the platform-standardization effort (#466): one token lifecycle for all platforms instead of per-platform token managers. Pure refactor — no tool changes, no behavior changes.
+
+### Changed
+- **New `platforms/_common/auth.py`** — `AsyncTokenManager` (cached token + expiry buffer + `asyncio.Lock` with double-checked locking, `invalidate()` / `refresh()` / `prime()` / `static()`), `TokenResult`, the shared `AuthError` base, and the `oauth2_client_credentials(...)` fetch-strategy factory. Extracted verbatim-in-spirit from the two proven in-tree implementations rather than designed fresh:
+  - **UXI** donated the OAuth2 client-credentials flow (time-based expiry with 60-second buffer, lock-serialized refresh, token POST on a fresh no-base_url client). `UXIClient` now wires `oauth2_client_credentials` into an `AsyncTokenManager` and drops its inline `_ensure_token` / `_fetch_token_locked` / `_invalidate_token` machinery.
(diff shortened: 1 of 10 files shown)


nowireless4u:
Addressed the suggestion in 166dd82: `static()` now rejects an empty token at construction, and `prime()` got the same guard — an empty cached token would have passed `_is_fresh()` (it only checks `is None`) and been sent as a blank credential, so both entry points now fail loudly the way `_fetch_locked()` already did. Two new unit tests cover the guards. Full suite re-run in Docker: 1788 passed, ruff/format/mypy clean.

Issue #469, opened 2026-06-11
ClearPass: pre-existing hand-written endpoints don't match the CPPM REST surface (preserved verbatim in the httpx rebuild)


nowireless4u:
## Context

During the pyclearpass→httpx rebuild (#465), every call site was checked against the SDK's path map. The rebuild preserved behavior byte-for-byte, which means a set of **pre-existing** hand-written paths that don't match the documented CPPM REST surface were carried over unchanged. They were almost certainly broken against a live box before the rebuild too. Separately, several old call sites invoked pyclearpass methods that don't exist in the SDK at all (instant `AttributeError`) — those are now fixed as a side effect of path-based calls, but the items below remain suspect.

## Suspect endpoints (preserved verbatim, need live probing + fixes)

**Guest actions (`manage_guests.py`)**
- `clearpass_send_guest_credentials`: POSTs `/guest/{id}/send-sms|send-email`; API has `POST /guest/{guest_id}/sendreceipt/sms|smtp` and expects a confirmation body, not `{}`.
- `clearpass_generate_guest_pass`: POSTs `/guest/{id}/pass` / `/receipt`; API has `GET /guest/{guest_id}/pass/{template_id}` / `/receipt/{template_id}`.
- `clearpass_process_sponsor_action`: POSTs `/guest/{id}/sponsor/approve|reject`; API has `POST /guest/{guest_id}/sponsor` with a `gsr_id` query param + body.

**Session control (`manage_sessions.py`)**
- Bulk disconnect/reauth POST `/session/disconnect` and `/session/reauthorize`; API surface is `/session-action/disconnect` and `/session-action/coa`.
- `clearpass_perform_coa` with `target_type="session_id"` issues `GET /session/{id}/reauthorize` — that returns reauthorization *templates*; it does not perform the CoA (the performing call is a POST with a body, or `/session-action/coa`).

**Server config (`manage_server_config.py`)**
- `/admin-user/name/{name}` — API route is `/admin-user/user-id/{user_id}`; no name route.
- `/file-backup-server/name/{name}` — API: `/file-backup-server/host-address/{host_address}`.
- `/attribute/name/{name}` — API: `/attribute/{entity_name}/name/{name}` (entity segment required).
- `PATCH /server/policy-manager-zones/{id}` and `/name/{name}` — API documents only PUT/DELETE.

**Local config (`manage_local_config.py`)**
[...]

Issue #415, opened 2026-06-03
Code-mode write confirmation is bypassable (confirmed=true) — and #414's "elicit() can't round-trip from the sandbox" premise is false


nowireless4u:
## Summary

The v3.2.3.10 (#414) fix for #413 is built on an incorrect assumption about FastMCP code mode, and as a result the code-mode write-confirmation gate is **advisory only** — an AI client can authorize any create/update/delete operation itself, with no human in the loop, by setting `confirmed=true`.

Two distinct findings:

1. **`confirmed=true` bypasses the confirmation gate entirely.** The inline write guards use `if action_type != "create" and not confirmed:` before invoking `elicitation_handler`. When the model sets `confirmed=true`, the elicitation path is never reached — no prompt, no server-side gate. The model decides on its own that approval was obtained. In code mode this is especially impactful because the client's *native* per-tool approval never fires either: the human approves one `execute()` block, and any number of nested writes run inside it via `call_tool` that the client never sees.

2. **#414's founding premise is false.** #414 defaults code-mode confirmation to `chat_confirm` on the stated grounds that *"an interactive `elicit()` cannot round-trip out of the sandbox."* That is not true (verified live, see below). `ctx.elicit()` round-trips out of the Monty sandbox identically to a top-level call, so a real human confirmation prompt **can** fire in code mode — and `_resolve_elicitation_mode()` hardcoding `chat_confirm` (it never returns `prompt`) permanently suppresses it.

## Observed behavior

An operator asked the assistant to bulk-remove a large set of Central config assignments. The assistant set `confirmed=true` on every write from the first call (treating a single chat "proceed" as blanket authorization), so no confirmation prompt ever surfaced and the server never gated any individual write. The human-in-the-loop guarantee collapsed to "the model asked once in chat and then self-approved N destructive operations."

## Verification (FastMCP 3.4.0)

A diagnostic harness (in-memory `Client` with elicitation/sampling/progress/log handlers → `execute` → `call_tool('probe')`, compared against a direct top-level call) showed the sandbox boundary is **transparent** to all four server→client channels:

| Capability | Direct (control) | Inside code-mode sandbox |
|---|---|---|
| `ctx.info()` | ok | ok |
| `ctx.report_progress()` | ok | ok |
| `ctx.elicit()` (capable client) | accepts | accepts (real prompt) |
| `ctx.elicit()` (no-capability client) | raises `McpError: Elicitation not supported` | raises identically |
[...]

Issue #126, opened 2026-04-17
ClearPass: Reliance on pyclearpass _send_request private method for list operations


nowireless4u:
## Problem

Most ClearPass read tools use `client._send_request("/resource?filter=...&sort=...", "get")` for list operations because the `pyclearpass` SDK lacks dedicated list methods for most resource types. The SDK only provides `get_*_by_id()` and `get_*_by_name()` methods.

`_send_request()` is a private method (prefixed with `_`) on `ClearPassAPILogin`. If pyclearpass changes its internals in a future release, all list operations would break.

## Affected Files

All 16 ClearPass read tool files that perform list operations use this pattern.

## Proposed Solutions

1. **Monitor pyclearpass releases** — if the SDK adds public list methods, migrate to them
2. **Pin pyclearpass version** — lock to `pyclearpass==1.0.8` instead of `>=1.0.0` to prevent breakage from SDK updates
3. **Wrap in our own helper** — create a `clearpass_list_request(client, path, filter, sort, offset, limit)` wrapper in `client.py` so there's a single place to update if the SDK changes

## Impact

Low risk currently — pyclearpass 1.0.8 is stable and the `_send_request` pattern is well-established. This is a future-proofing concern, not an immediate issue.

PR #470, opened 2026-06-11
feat(clearpass)!: rebuild ClearPass on async httpx — remove pyclearpass SDK + full capability migration (v3.3.12.0)

### LINKED ISSUE #441: ClearPass client performs blocking HTTP work from async tool paths
## Bug description
ClearPass tools are exposed through async MCP paths, but both OAuth refresh and pyclearpass API calls are synchronous.

## Evidence
- `src/hpe_networking_mcp/platforms/clearpass/client.py:59-74` refreshes OAuth with synchronous `httpx.post(...)`.
- `src/hpe_networking_mcp/platforms/clearpass/client.py:189-234` exposes async `get_clearpass_session(...)`, then calls `token_manager.get_token()` synchronously.
- `src/hpe_networking_mcp/platforms/clearpass/client.py:125-153` wraps pyclearpass `_send_request`, which remains synchronous.
- `src/hpe_networking_mcp/platforms/clearpass/client.py:183-186` constructs pyclearpass clients using the synchronous token manager path.

## Why it matters
Slow ClearPass OAuth/API calls can block the MCP event loop and starve unrelated platform tools. This is especially risky during token expiry or controller/network problems.

## Suggested fix
Wrap pyclearpass calls and token refreshes with `asyncio.to_thread`, or replace the ClearPass transport with async HTTP.

Add a regression test where a slow ClearPass request does not block concurrent async tool work.

### LINKED ISSUE #465: ClearPass: replace pyclearpass with a template-conformant async httpx client (bundle with capability migration)

nowireless4u:
## What

Closes #441 (blocking HTTP from async paths). Closes #465. Step 3 of the platform-standardization effort (References #466 — release stays gated until all platforms are on the shared auth system).

The bundled per-file sweep #465 called for: pyclearpass removal + `capability=` migration in one pass over the ~36 affected files, so each file was touched once. Net diff is **negative** (−72 lines) despite adding a full client.

### The client
- **`ClearPassClient`** preserves the pyclearpass response contract exactly: decoded JSON with raw-text fallback, raw bytes for non-JSON accepts, and **error bodies returned, never raised** (tools inspect `{"status": 403, ...}` dicts — verified against `ClearPassAPILogin._send_request`'s source line-by-line). Wire parity includes pyclearpass's quirks: `""`-valued query params and top-level body keys are dropped, dict-valued params travel as compact JSON (`filter` semantics).
- Token lifecycle on the shared `AsyncTokenManager` (JSON `POST /oauth`, same error messages as the old `ClearPassTokenManager`). Expiry is now tracked proactively from `expires_in` (8 h) — the post-expiry 403 storm the old code worked around **structurally cannot happen**. The 401/403 refresh-once-and-replay that previously lived in a class-level `_send_request` monkey-patch is now first-class client behavior.

### Conversion mechanics (for reviewer confidence)
- The SDK's 796 methods were mapped **mechanically** (AST/regex extraction of verb + path template + param placement from the SDK source) rather than hand-transcribed; every converted call site was checked against that map, and divergences were flagged, not silently "fixed".
- The `clearpass_get` chokepoint (issue #126's isolation) converted ~70 read sites in one function.
- Post-sweep gates: compile pass, un-awaited-coroutine AST audit (clean), no-pyclearpass guard test (permanent), full suite, live CPPM verification.

### Capability migration (3/N of the annotation sequence)
- `_registry.py` on the shared `make_tool_decorator` factory; all 142 tools classified per `docs/tool-annotation-rubric.md`.
- Gating is **byte-identical**: `Capability.WRITE_DELETE` (and `OPERATIONAL`/`WRITE` with explicit gating where flagged) derives exactly the old hand-written `clearpass_write_delete` enable tag. The existing dynamic-mode test asserting that tag passes unchanged — a free regression check on the classification.
- **Deviation from #465's text, deliberate:** inline confirmation machinery is NOT removed here. The universal `requires_confirmation` gate (#415/#416 step 3) doesn't exist yet — removing inline confirms now would leave writes unguarded. They go in the gate PR.

### Bugs surfaced
- **Fixed as a side effect**: several tools called pyclearpass methods that don't exist in the SDK (e.g. `get_admin_user_by_admin_user_id`) — `AttributeError` on every invocation, i.e. those code paths never worked.
- **Preserved + catalogued**: ~14 hand-written endpoints that don't match the CPPM REST surface (guest send/sponsor actions, bulk session disconnect/CoA routing, several `/name/{name}` routes, AD-domain path order, extension log path). All preserved byte-for-byte per the behavior-preserving rule and filed as #469 for live probing.

## Tests & verification
- New `tests/unit/test_clearpass_client.py`: response contract (error-bodies-returned, decode rules, parity quirks), auth-retry flow (403→refresh→replay, no loops, refresh-failure fallback), token fetch contract, lazy construction, plus the permanent no-pyclearpass import guard.
- Full suite in Docker: **1818 passed, 1 skipped**; ruff + format + mypy + bandit clean.
- **Live-verified against a real CPPM** through the new client: OAuth2 token fetch, NAD list read (HAL envelope intact), `cppm-version` read, and the 404-error-body-as-dict contract.

## Platform-standardization status after this PR

Shared auth adopters: UXI, Apstra, `_template`, Central, **ClearPass**. Capability-migrated: Axis, AOS8, **ClearPass**. Both vendor SDKs are now gone from the dependency tree. Remaining: GreenLake, AOS8 (client conformance), Mist, Axis adoptions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.3.12.0] - 2026-06-11
+
+**Minor — ClearPass rebuilt on async httpx; pyclearpass SDK removed; full `capability=` migration.** Closes #441, #465. Step 3 of the platform-standardization effort (#466). No release/tag until the auth migration completes across all platforms.
+
+### Changed
+- **New `ClearPassClient`** (`platforms/clearpass/client.py`) replaces the pyclearpass SDK + `ClearPassTokenManager` + the `_send_request` monkey-patch. The response contract is pyclearpass-identical: decoded JSON with raw-text fallback, raw bytes for non-JSON accepts, and **error bodies returned, never raised**. Wire parity preserved down to pyclearpass's quirks: `""`-valued query params and top-level body keys are dropped, dict-valued params are compact-JSON encoded. Token lifecycle runs through the shared `AsyncTokenManager` (JSON `POST /oauth`, proactive 8-hour expiry tracking — the 403 storm after a token aged out structurally can't happen anymore); 401/403 auth-failure bodies trigger one refresh + replay, now first-class instead of monkey-patched.
+- **Every ClearPass call is now genuinely async** (closes #441): ~250 call sites across 36 files (35 tool modules + `site_health_check`) converted from sync SDK calls to `await client.request(...)`; the `clearpass_get` chokepoint (issue #126's isolation paid off) converted once for ~70 read sites; `asyncio.to_thread` band-aids removed.
+- **Full capability migration** (3/N of the annotation sequence): `_registry.py` on the shared `make_tool_decorator` factory; all 142 tools classified with `capability=` per the rubric (READ for reads, WRITE_DELETE for manage surfaces, OPERATIONAL+`enable_gated` for disruptive non-persistent actions like session disconnect/CoA, guest credential sends, and service start/stop). The derived `clearpass_write_delete` enable tag is byte-identical to the old hand-written tags — write gating is unchanged, and inline confirmation logic is untouched (the universal `requires_confirmation` gate lands in a later PR per the #415/#416 sequence).
+- `pyclearpass` removed from dependencies; permanent guard test (`TestNoPyclearpassAnywhere`) fails on any reappearing import. `server.py` lifespan now builds one lazy `clearpass_client` (replacing the `clearpass_config` + `clearpass_token_manager` pair); the health probe uses the client.
+
+### Fixed
+- **Several tools called pyclearpass methods that don't exist in the SDK** (e.g. `get_admin_user_by_admin_user_id`, `get_config_service_by_config_service_id`) — instant `AttributeError` on every invocation. Path-based calls fix them as a side effect.
+- Pre-existing hand-written endpoints that don't match the CPPM REST surface were **preserved verbatim** (behavior-preserving rebuild) and catalogued in #469 for live probing + fixes.
+
+### Tests
+- New `tests/unit/test_clearpass_client.py` (response contract incl. error-bodies-returned, auth-retry flow, token fetch, pyclearpass parity quirks) + updated health/visualizer/utils test fixtures to async mocks.
+- Full suite **1818 passed** in Docker; ruff + format + mypy + bandit clean. **Live-verified against a real CPPM**: token fetch, NAD/version reads (HAL envelope intact), and the 404-as-dict error contract.
+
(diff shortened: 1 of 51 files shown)


nowireless4u:
Addressed the test suggestion in c7763a6: added `TestBasePathJoining` pinning that a server URL with a path component (`https://host:443/api`) produces requests under `/api/...` for both normal API calls and the `/oauth` token fetch. (This behavior was already proven by the live CPPM verification — the real server URL carries `/api` — but now it's pinned against regression.) The site_health_check params= consolidation is noted for the #469 follow-up, where those call sites get touched anyway. 18/18 client tests green, style clean.

Issue #440, opened 2026-06-07
GreenLake token acquisition blocks the async server runtime


nowireless4u:
## Bug description
GreenLake tools are async, but token acquisition and refresh use synchronous `httpx.Client` calls from the async request path and startup path.

## Evidence
- `src/hpe_networking_mcp/platforms/greenlake/client.py:33-49` defines async `get()` but calls `self._get_auth_headers()` before awaiting the async HTTP request.
- `src/hpe_networking_mcp/platforms/greenlake/client.py:91-95` calls `self.token_manager.get_auth_headers()` synchronously.
- `src/hpe_networking_mcp/platforms/greenlake/auth.py:107-116` uses synchronous `httpx.Client().post(...)` to fetch OAuth tokens.
- `src/hpe_networking_mcp/platforms/greenlake/auth.py:203-206` fetches a token in `TokenManager.__init__` when no initial token exists.
- `src/hpe_networking_mcp/server.py:86-99` constructs `TokenManager` during async lifespan startup.

## Why it matters
A slow token endpoint, network timeout, or token refresh can block the whole MCP event loop for up to the token request timeout. That can stall unrelated tools and make the server look hung.

## Suggested fix
Make GreenLake token management async end-to-end with `httpx.AsyncClient` and an async lock, or isolate synchronous token fetch/refresh with `asyncio.to_thread`.

Also avoid doing blocking token acquisition in `TokenManager.__init__`; prefer lazy async initialization on first request or an explicit startup probe that cannot block the event loop.

nowireless4u:
Fixed in #463 (merged to main). Released in v3.3.10.1.

PR #471, opened 2026-06-12
refactor(auth): final shared-auth adoptions — GreenLake, AOS8, Mist, Axis (v3.3.12.1)

nowireless4u:
## What

The last four platforms adopt the shared `AsyncTokenManager` — **the auth migration is complete**. All 8 platforms (+ `_template`) now run token lifecycle through one primitive; zero platform-owned token managers remain. Net diff: **−469 lines**.

References #466 (completes its auth portion; the registry-shim + client-conformance items remain open there).

### Per platform
- **GreenLake** — `greenlake/auth.py` (290-line sync `OAuth2Provider`/`TokenManager`) **deleted**. `make_token_manager()` configures the shared `oauth2_client_credentials` strategy: workspace-scoped token URL, form-body credentials (`client_secret_post`, what the old provider sent), and the old **300-second expiry buffer preserved**. Token acquisition is natively async — the #440 `to_thread` workaround and the eager blocking fetch in the lifespan are gone. Health probe updated.
- **AOS8** — UIDARUBA login-session adopts the manager: `_login()` is the fetch strategy (same validation + error messages), 401 → `refresh()` retry preserved, and the platform's quirk — **SESSION-cookie rotation on every response** — syncs back via `manager.prime()`. Credential *attachment* (cookie + query param) stays in the client, exactly as the #466 design intended. `AOS8AuthError` subclasses the shared `AuthError`.
- **Mist / Axis** — static tokens routed through `AsyncTokenManager.static()`: uniform auth surface plus the empty-token construction guard from the #467 review. Axis keeps its JWT expiry-days advisory; `AxisAuthError` subclasses `AuthError`.

### Verification
- **Live**: GreenLake token fetched through the new manager against the real workspace endpoint (expiry tracked from `expires_in`); AOS8 real controller login + `show version` through the manager-backed flow (cookie rotation path exercised).
- Full Docker suite: **1802 passed, 1 skipped**; ruff + format + mypy + bandit clean.
- `test_greenlake_auth.py` (22 tests for the deleted module) replaced by focused wiring tests; lifecycle behavior is covered once in `test_common_auth.py` instead of per-platform.

## Release gate

Per the standing decision, no release has been tagged since v3.3.10.1 while the migration was in flight. **This PR completes the migration** — once merged, v3.3.12.1 is the tag/release candidate covering the whole span: the auth primitive (#467), the Central rebuild (#468), the ClearPass rebuild (#470), and these adoptions.

## End state

| Platform | Auth | Vendor SDK |
|---|---|---|
| UXI, Apstra | shared manager (donors, #467) | — |
| Central | shared manager, `client_secret_basic` (#468) | pycentral removed |
| ClearPass | shared manager, JSON `/oauth` fetch (#470) | pyclearpass removed |
| GreenLake | shared manager, workspace OAuth2 (this PR) | — |
| AOS8 | shared manager, login-session + cookie rotation (this PR) | — |
| Mist, Axis | shared manager, static (this PR) | — |


nowireless4u:
Addressed in dd38c27: the wiring test now also asserts `client_secret` is in the form body and that no `Authorization` header is sent on the token request — the `client_secret_post` contract is fully pinned. Tests green.

PR #475, opened 2026-06-12
fix(clearpass): repair all 14 broken endpoints from the #469 catalog (v3.3.12.5)

### LINKED ISSUE #469: ClearPass: pre-existing hand-written endpoints don't match the CPPM REST surface (preserved verbatim in the httpx rebuild)

nowireless4u:
## What

Closes #469. Every endpoint catalogued there has now been **live-probed against a real CPPM** with bogus identifiers — both directions:

1. **Before fixing**: every suspect path confirmed genuinely broken — HTTP 405 with empty HTML body (route/method unsupported), page-level 404, or a 422 proving the `name/` segment was being parsed as an ID. Zero false alarms in the catalog.
2. **After fixing**: all 19 corrected routes re-probed — every one responds with object-level validation (404 "Object not found" / 422 / Guest-module responses), zero still-broken.

## Fixes by area

- **Guest actions**: sendreceipt routes with the required `confirm` flag (email→`smtp` channel); pass/receipt **rendering** is a GET with a template ID (tool reclassified READ + new required `template_id`); sponsor approval/rejection is one POST with the sponsor-link `token`/`register_token`, rejection via `register_reject`.
- **Session control**: bulk + selector disconnects on the `/session-action/` surface (selector variants for mac/username/ip; bulk filter properly wrapped). **The headline fix: `clearpass_perform_coa` by session ID previously issued a GET that returns reauthorization templates — it never performed a CoA.** Now `POST /session/{id}/reauthorize` with `confirm_reauthorize`; other targets use `/session-action/coa[...]` with the API-mandated `enforcement_profile` (new required param there, validated with a clear 400).
- **Server/local config**: `user-id`/`host-address`/`server-name` selector routes; entity-scoped attribute names (new validated `entity_name` param); PATCH→PUT where the API only documents PUT; action-prefix AD-domain routes (`/ad-domain/join/{uuid}`); `/service-parameter/{uuid}/{id}`.
- **Integrations**: `/extension/instance/{id}/log` (the hyphenated form 404s).

## Why the signature additions are safe

Several of these tools called pyclearpass methods that don't exist (pre-rebuild `AttributeError`) or hand-written paths that 405 — **none of them ever worked against a live CPPM**, so the new required params (`template_id`, sponsor tokens, `enforcement_profile`, `entity_name`) cannot break a working caller. Each is validated with a clear 400 + guidance.

## Tests

- New `tests/unit/test_clearpass_endpoint_routes.py` — 12 tests pinning verb + path + load-bearing body fields for every corrected route (including the wrapped-filter and `register_reject` details).
- Full suite: **1814 passed, 1 skipped**; ruff + format + mypy + bandit clean.
- docs/TOOLS.md updated for the changed parameter surfaces.


nowireless4u:
Two responses to the review:

**On the Critical finding** — the committed assertion already is `{"token": "t", "register_token": "rt", "register_reject": True}` (tests/unit/test_clearpass_endpoint_routes.py, sponsor-reject test). It looks like the diff view rendered the one-character literal `"t"` as `"***"` — a secret-masking artifact on the review side. Two independent confirmations: `git show` of the branch shows `"token": "t"`, and the **Tests CI check passes**, which would be impossible if the assertion expected `"***"` against a body built from `token="t"`. No code change made for this item.

**On the Suggestion** — taken: `dd2de8f` adds `test_generate_receipt_renders_receipt_route` pinning `GET /guest/{id}/receipt/{template_id}`, so both render paths are covered. 13/13 route tests green, style clean.

nowireless4u:
Both findings addressed:

**docker-compose.yml — you were right, good catch.** My style-fixup commit (2b2ea15) used `git add -u` and swept the maintainer's local runtime compose file (which legitimately carries enabled write flags for his deployment) into the PR. Reverted in c8b8e23 — the file's net diff against main is now zero. The shipped defaults remain `false`. (Note: the interim values were deployment toggles, not secrets — credentials live in Docker secrets files, never in compose.)

**Sponsor-token assertion — standing by the previous response.** The branch content is `"token": "t"` (verifiable via the Files view on this PR, not the unified diff, which appears to mask the one-character string literal as `***` in your pipeline). The Tests CI check has passed on every push — impossible if the assertion expected `***`. No change made.

13/13 route tests green; awaiting the CI re-run on c8b8e23.

nowireless4u:
Round-4 token finding: the assertion has asserted the exact passed value on every push (Tests CI has never failed), but your diff pipeline masks string literals in token-named positions, which kept rendering the expectation as `***` on your side. Rather than keep arguing with the masker, 529d050 binds the fixture values to variables and asserts the dict against those variables — no literals left to mask, identical coverage. 13/13 route tests green.

Issue #436, opened 2026-06-07
Generated Mist write tools bypass elicitation confirmation


nowireless4u:
## Bug description
Generated Mist write/delete tools are enabled by tag but do not call the confirmation helper before sending POST/PUT/PATCH/DELETE requests.

## Evidence
- `src/hpe_networking_mcp/middleware/elicitation.py:81-83` enables tools tagged `mist_write` / `mist_write_delete` when Mist writes are enabled.
- `src/hpe_networking_mcp/platforms/mist/tools/sites_psks.py:26-44` registers `mist_create_site_psk` as `mist_write` and immediately calls `mist_request("POST", ...)`.
- `src/hpe_networking_mcp/platforms/mist/tools/sites_psks.py:47-65` registers `mist_delete_site_psk` as `mist_write_delete` and immediately calls `mist_request("DELETE", ...)`.
- `src/hpe_networking_mcp/platforms/_common/meta_tools.py:473-474` dispatches dynamic/code-mode calls with `return await spec.func(ctx, **coerced)`, so the meta-tool path does not add a missing confirmation layer.

## Why it matters
Once `ENABLE_MIST_WRITE_TOOLS=true`, generated Mist write tools can mutate or delete Mist state without the user confirmation contract used elsewhere. That includes direct tool calls and dynamic/code-mode `mist_invoke_tool` calls.

## Suggested fix
Add a centralized confirmation wrapper for generated Mist mutations, preferably in the Mist request/registry layer or generator template. Gate all non-GET methods or all tools tagged `mist_write` / `mist_write_delete` before the HTTP request is sent.

Add regression tests for both:
1. Direct generated Mist write tool call.
2. `mist_invoke_tool` dispatch to a generated Mist write tool.

Both should return `confirmation_required` before any HTTP call when elicitation is not disabled.

Issue #416, opened 2026-06-03
Central MRT + alert write tools execute destructive operations with NO confirmation gate


nowireless4u:
## Summary

A number of Aruba Central write tools — concentrated in the MRT (Monitoring / Reporting / Troubleshooting) bulk-import surface and `alert_configs` — perform `POST`/`PUT`/`PATCH`/`DELETE` operations **without ever invoking the confirmation gate** (`elicitation_handler` / `confirm_write`). They carry the `central_write_delete` tag (so they're hidden unless `ENABLE_CENTRAL_WRITE_TOOLS=true`), but once enabled they execute destructive operations with no per-operation user confirmation on any client type.

This is distinct from the `confirmed=true` bypass tracked separately — these tools have **no gate at all**, not a bypassable one.

## Evidence

Audited the central tool modules for any non-GET `retry_central_command(...)` call preceded by a confirmation gate. The following files contain **zero** `elicitation_handler`/`confirm_write` calls despite multiple non-GET writes:

| File | gate calls | non-GET write call-sites |
|---|---|---|
| `alert_configs.py` | 0 | 4 |
| `mrt_topology.py` (incl. `central_delete_device`) | 0 | 4 |
| `mrt_reporting.py` (`central_update_report` — docstring *claims* it confirms) | 0 | 3 |
| `mrt_webhooks.py` (HMAC secrets in payloads) | 0 | 7 |
| `mrt_sitemaps.py` | 0 | 21 |
| `mrt_services.py` | 0 | 9 |

Representative destructive operations reachable with no confirmation: delete a device from inventory, create/rotate/delete a webhook, reset an alert config to inherited, create/update/delete floors and buildings, update a report definition.

## Not affected

The ~197 generated config-model tools (`central_manage_*`) are **fine** — they delegate writes to `_manage_resource()` (`security_policy.py`), which gates. Hand-curated tools (`config_assignments`, `roles`, `sites`, `configuration`, `gateway_clusters`, `wlan_profiles`, `security_policy*`) gate correctly.

## Root cause

The MRT surface was bulk-imported (v3.1.2.0) and the per-tool confirmation pattern was not applied to the generated write tools. This is the structural weakness of per-tool gating: every new write tool must *remember* to call the gate, and a bulk import didn't.

## Related annotation-hygiene defect

[...]

PR #477, opened 2026-06-12
feat(gating)!: universal confirmation gate — destructive-tool confirmation becomes structural (v3.3.13.0)

### LINKED ISSUE #415: Code-mode write confirmation is bypassable (confirmed=true) — and #414's "elicit() can't round-trip from the sandbox" premise is false

nowireless4u:
## What

**Closes #415, #416, #436.** The capstone of the elicitation-gating effort: one confirmation gate at the `<platform>_invoke_tool` dispatch chokepoint — the path BOTH code mode and dynamic mode use to reach all 1962 tools — replaces ~75 hand-written per-tool confirmation blocks.

### The gate (`confirm_gated_invoke`)
- Keyed ONLY on the `requires_confirmation` tag, which every tool now derives automatically from its `capability=` classification (the 8-platform migration that preceded this PR). Never inferred from hints. **Fail-closed**: an unclassified tool is gated by omission.
- Decision sequence:
  1. `DISABLE_ELICITATION=true` → auto-accept (operator opt-out, unchanged).
  2. A **real `ctx.elicit()` prompt is attempted first** — it round-trips transparently from the code-mode sandbox (verified live 2026-06-03; the contrary premise from #414 was false). The human's accept/decline/cancel always wins.
  3. **`confirmed=true` is ignored while a prompt is available** — this is the #415 fix. The flag carries authority only when the client genuinely cannot present a prompt (elicit raises), as the confirm-in-chat fallback.

### What this closes
- **#415** — code-mode `confirmed=true` self-authorization: dead. An AI cannot skip a working prompt.
- **#416** — the ~20 Central MRT/alert destructive tools that had NO confirmation pathway: now gated structurally (their capability classification landed in #473).
- **#436** — all 558 generated Mist write tools that bypassed elicitation: now gated (classification landed in #474).

### Inline machinery removed
~75 confirm constructs across 34 files (ClearPass 56 alone), their local `_confirm_*` helpers and imports, and the dead `_build_patch_diff` elicitation-diff renderer. `confirmed` parameters REMAIN on tools — the gate consumes them for the fallback — with descriptions updated to the fallback-only semantics. Keeping the inline blocks alongside the gate would double-prompt every write.

### Statics
`manage_wlan_profile` (the one direct-call write-tagged static that bypasses `_invoke_tool`) was audited: it performs **no mutations** — its only API call is a read; everything else returns workflow guidance. No gate needed; enable-gating retained.

## Tests
- `test_universal_confirmation_gate.py` — 9 unit tests pinning the decision sequence, including the load-bearing one: *confirmed=true ignored while a prompt is available*.
- `test_gate_end_to_end.py` — 4 tests driving the REAL dispatch path over an in-memory MCP session with a client-side elicitation handler: prompt→accept→dispatch; prompt→decline blocks despite `confirmed=true`; READ tools never prompt; a no-elicitation client gets `confirmation_required` then succeeds on the `confirmed=true` retry.
- Obsolete inline-confirm tests across aos8/uxi/central/clearpass suites updated or retired with pointers to the gate tests.
- Full suite in Docker: **1826 passed, 1 skipped**; ruff + format + mypy + bandit clean.
- INSTRUCTIONS.md confirmation guidance rewritten for the new contract.

## What remains on #466 after this
Only the cleanup PR (retire now-unused `confirm_write`/`elicitation_handler` once nothing references them, plus any leftover duplicate code). The elicitation-gating effort's behavioral work is complete with this merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.3.13.0] - 2026-06-12
+
+**Minor — the universal confirmation gate. Closes #415, #416, #436.** Destructive-tool confirmation is now structural and non-bypassable: one gate at the `<platform>_invoke_tool` dispatch chokepoint (the path BOTH code mode and dynamic mode use) replaces ~75 hand-written per-tool confirmation blocks. The capstone of the elicitation-gating effort (#415/#416 design → capability rubric → 8 platform migrations → this).
+
+### Added
+- **`confirm_gated_invoke`** (`middleware/elicitation.py`) + its call in `_invoke_tool`: any tool carrying the `requires_confirmation` tag (derived automatically from its `capability=` classification) is confirmed BEFORE dispatch. **Fail-closed**: an unclassified tool is treated as gated.
+- The decision sequence: `DISABLE_ELICITATION=true` → auto-accept; otherwise a **real `ctx.elicit()` prompt is attempted first** — it round-trips transparently from the code-mode sandbox (verified live 2026-06-03; #414's contrary premise was false). **`confirmed=true` is ignored while a prompt is available** — the human's accept/decline/cancel always wins (this closes the #415 self-authorization hole). Only when the client cannot present a prompt (elicit raises) does `confirmed=true` carry authority as the confirm-in-chat fallback.
+- The ~20 formerly-ungated Central MRT/alert destructive tools (#416) and all 558 generated Mist write tools (#436) are now confirmed through the gate — no per-tool opt-in required, ever again.
+
+### Removed
+- All inline per-tool confirmation machinery: ~75 `confirm_write`/local-helper constructs across 34 files (ClearPass 56, plus Central scope/config/security/WLAN, Apstra, UXI, Axis, AOS8), their local `_confirm_*` helpers, and the dead `_build_patch_diff` elicitation-diff renderer. `confirmed` tool parameters REMAIN (the gate consumes them for the fallback path); their descriptions now state the fallback-only semantics.
+- `manage_wlan_profile` (direct-call static) verified mutation-free — it returns workflow guidance only and performs no writes, so it carries no confirmation (enable-gating retained).
+
+### Tests
+- `test_universal_confirmation_gate.py` (9 tests): decision sequence incl. the confirmed-ignored-while-prompt-available pin and fail-closed predicate.
+- `test_gate_end_to_end.py` (4 tests): a REAL in-memory MCP session with an elicitation handler driving `_invoke_tool` — prompt fires and dispatches on accept, decline blocks even with `confirmed=true`, reads never prompt, and a no-elicitation client gets `confirmation_required` then succeeds via the `confirmed=true` retry.
+- Obsolete inline-confirm tests updated/retired across aos8/uxi/central/clearpass suites. Full suite **1826 passed**; ruff + format + mypy + bandit clean.
+
 ## [3.3.12.5] - 2026-06-11
 
 **Patch — all 14 broken ClearPass endpoints from #469 live-probed and fixed.** Closes #469. Every suspect path was confirmed broken against a live CPPM (HTTP 405 / page-404 / wrong-ID parse), every corrected route verified live with bogus identifiers before and after the fix, and regression tests now pin verb + path + load-bearing body fields.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "hpe-networking-mcp"
-version = "3.3.12.5"
+version = "3.3.13.0"
 description = "Unofficial, community-driven MCP server for Juniper Mist, Aruba Central, HPE GreenLake, and ClearPass. Not affiliated with HPE, Aruba, or Juniper."
 readme = "README.md"
 license = "MIT"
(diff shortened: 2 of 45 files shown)


nowireless4u:
All three findings addressed in 0f74eab:

**Critical (param visibility)** — implemented as suggested: prompts now read `Confirm: <platform> tool '<name>' (<summary>)` + `Params: {...}` with a sanitized summary — sensitive keys (`password`/`secret`/`token`/`psk`/…) redacted at any nesting depth, the bookkeeping `confirmed` flag dropped, output capped at 300 chars. The chat-fallback message carries the same summary so in-chat confirmation is equally informed. Tests pin target-visible + secret-redacted + length-capped.

**Warning (broad except)** — narrowed exactly as suggested: probed FastMCP live; the no-capability signal is `McpError('Elicitation not supported')`, now the ONLY trigger for the `confirmed=true` fallback. Any other exception fails closed with a distinct `confirmation_unavailable` status that does NOT honor `confirmed=true` (new test pins handler-crash + confirmed=true → blocked). The e2e no-handler test still passes, confirming the narrow catch matches real client behavior.

**Suggestion (stale INSTRUCTIONS sentence)** — rewritten; the doc now consistently states prompts round-trip from code mode and the guard dict is the no-capability path only.

Full suite green; ruff/format/mypy/bandit clean.

nowireless4u:
Round-3 findings addressed in ea16906:

**McpError specificity** — the fallback now requires the exact no-capability signal: `code == -32601` (METHOD_NOT_FOUND) or the verified `'Elicitation not supported'` message. Every other `McpError` (transport failure, malformed response, framework error) fails closed with `confirmation_unavailable` and does NOT honor `confirmed=true` — identical handling to non-MCP failures. New test pins a transport-flavored `McpError` + `confirmed=true` → blocked.

**Redaction breadth** — keys are now normalized (lowercased, separators stripped) and matched against sensitive suffixes (`password`/`secret`/`token`/`key`/`psk`/`passphrase`/`community`/`credential`), so `api_key`, `apiKey`, `api-key`, `clientSecret`, `refreshToken`, `private_key`, and compounds like `radiusSharedSecret` all redact. Over-redaction is the chosen failure mode. New test pins five camelCase/compound leak candidates redacted while non-sensitive values stay visible.

**On the warning** (param summary in AI-visible fallback output) — kept, as you anticipated: with the hardened normalized redaction the summary is the informed-consent payload for the in-chat confirmation, and the same scrubber covers both surfaces.

Full suite 1831 green; ruff/format/mypy/bandit clean through the hard-gated battery.

nowireless4u:
The round-4 suggestion is the same diff-masker artifact from PR #475: the file passes `"apiKey": "LEAK1"` (verifiable in the Files view; the test would also be vacuous-pass-proof in CI if the literal were `***`). Regardless, eaa8cce binds the leak markers to variables so your pipeline renders call and assertion identically — no literals left in secret-named positions. 17/17 gate tests green. With the verdict at no-blocking, this should be ready.

Issue #493, opened 2026-06-19
Make skills-first discovery structural (return skill matches as result data) instead of prose-gated


nowireless4u:
## Context

Skills-first behavior is currently enforced only through **untrusted channels**: the server `instructions` blob (`INSTRUCTIONS.md`) and prepended tool-description prose (`_SKILLS_FIRST_GATE` on `search`/`tags`/`get_schema`, plus the `execute` PREREQUISITE line — `src/hpe_networking_mcp/server.py:632`).

Frontier models (Claude, GPT) do not reliably honor server-provided instructions or procedural directives embedded in tool descriptions. Small local models appear to honor it only because they are less injection-hardened and read descriptions more literally. So the gate is fragile by design and cannot be relied on across the model spectrum.

## Proposal (structural)

Surface skill matches through the **trusted channel** — tool *result* data. When `search` (or another discovery call) matches a bundled skill, include the match(es) as a structured field in the **result**:

```json
{
  "skills": [{"name": "...", "title": "...", "load_with": "skills_load(name='...')"}],
  "tools":  [ ... ]
}
```

The suggestion then arrives as an observation the model acts on, not an instruction it must trust. Skills stay discoverable across all models without depending on prose.

## Relation

Companion to the `Unknown tool` issue: that fix explicitly must **not** carry skills-first prose; this issue provides the structural mechanism instead.

---
Surfaced during external small-model testing in LM Studio (glm-4.7-flash; also reproduced on Qwen3 4B Instruct). Part of the code-mode small-model hardening set.

nowireless4u:
🚧 In progress — being worked as part of the code-mode small-model hardening effort (#488–#493), implemented sequentially (first one landed as #495). Flagging to avoid duplicate effort; thanks for any interest, and feel free to comment here with ideas.

PR #494, opened 2026-06-19
Accept platform in code-mode search discovery

### LINKED ISSUE #488: Code mode: `search` discovery tool hard-rejects unknown args (e.g. `platform`) → search never runs
## Symptom

In code mode, the `search` discovery tool rejects any argument beyond its declared params. A model searching for Central sites naturally calls:

```
search(query="sites list central", platform="central")
```

and gets:

> Failed to parse arguments for tool "search": params is not allowed to have the additional property "platform"

The call is rejected wholesale — **search never executes**. Because `search` is the discovery entry point, this blocks everything downstream. Reproduced across two small local models.

## Root cause

The code-mode `Search` tool (`fastmcp.experimental.transforms.code_mode.Search`, wrapped by `_HpeSearch` in `src/hpe_networking_mcp/server.py:594`) has a strict input schema (`additionalProperties: false`, only `query`/`detail`). Small models routinely pass a `platform` arg ("search central for X"), which is a perfectly reasonable intent but fails validation.

## Proposed fix (structural)

Make `search` tolerant of extra arguments — relax `additionalProperties` and/or accept an optional `platform` (and other common) param, folding it into the query or ignoring it. Verify `Search.fn`'s signature first so a relaxed schema doesn't just move the failure to a downstream `TypeError`. The same hardening likely applies to `tags`/`get_schema` (same strict-schema family).

This is a schema/behavior fix — model-agnostic, no reliance on guidance text.

---
Surfaced during external small-model testing in LM Studio (glm-4.7-flash; also reproduced on Qwen3 4B Instruct). One of a set of code-mode hardening issues for small local models; all fixes are structural (schema/behavior/result-data), since server `instructions` and tool-description prose are not reliably honored across models.

gaoflow:
Closes #488

## Summary
- Add an optional `platform` argument to the top-level code-mode `search` discovery tool.
- Fold `platform` into the existing tag filter before delegating to FastMCP search, so calls like `search(query="sites list", platform="central")` validate and stay scoped to Central tools.
- Add a regression test covering both the generated schema and platform-scoped search behavior.

## Tests
- `uv run --frozen ruff check .`
- `uv run --frozen ruff format --check .`
- `uv run --frozen mypy src/ --ignore-missing-imports`
- `uv run --frozen pytest tests/ -q` (1847 passed, 12 skipped)
- `git diff --check`

Docker compose CI was not run locally because the Docker daemon is unavailable at `/Users/vinzzy/.orbstack/run/docker.sock`.

AI assistance was used under my direction.

### DIFF
--- a/src/hpe_networking_mcp/server.py
+++ b/src/hpe_networking_mcp/server.py
@@ -2,7 +2,7 @@
 
 from contextlib import asynccontextmanager
 from pathlib import Path
-from typing import Any
+from typing import Annotated, Any
 
 from fastmcp import FastMCP
 from loguru import logger
@@ -575,7 +575,10 @@ def _register_code_mode(mcp: FastMCP, max_duration_secs: float = 30.0) -> None:
             GetTags,
             MontySandboxProvider,
             Search,
+            ToolDetailLevel,
         )
+        from fastmcp.server.context import Context
+        from fastmcp.tools import Tool
         from pydantic_monty import ResourceLimits
     except ImportError as e:
         logger.error(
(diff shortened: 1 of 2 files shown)


nowireless4u:
Thank you for this, @gaoflow — and for the careful work: folding `platform` into the existing tag filter is exactly the right instinct, and the regression test is appreciated.

I want to be transparent that a fix for #488 was already in flight on my side as part of a broader code-mode small-model hardening effort (#488–#493), now up as #495. After comparing the two, I'm going to land #495 because it covers a superset of this change:

- it hardens **all three** discovery tools (`search`, `tags`, `get_schema`), not just `search`;
- it drops **any** unknown argument rather than only accepting `platform`, so other stray args from small models don't dead-end either; and
- it tolerates `platform` as a string, a list, **or** a malformed scalar (e.g. `123`) — a non-string value would still fail validation under the explicit `str | None` param here.

None of that takes away from your diagnosis or fix — you landed on the same root cause and the same core remedy independently. I'm going to close this one in favor of #495 as a duplicate-in-flight, with full credit to you for the work.

I'd genuinely welcome more contributions from you down the line — thanks again for jumping in. 🙏

PR #495, opened 2026-06-19
fix(code-mode): tolerate extra args on search/tags/get_schema (#488)

### LINKED ISSUE #488: Code mode: `search` discovery tool hard-rejects unknown args (e.g. `platform`) → search never runs

nowireless4u:
## Summary

Closes #488 — the first of the code-mode small-model hardening set (#488–#493).

In code mode, the `search` discovery tool hard-rejected any undeclared argument. A small model searching for Central sites naturally calls:

```
search(query="sites list central", platform="central")
```

and got `Unexpected keyword argument` — the **whole call** was rejected, so `search` never ran. Since `search` is the discovery entry point, this dead-ended everything downstream (give-up loops, and in the worst case confabulated data).

## Root cause

fastmcp validates discovery-tool arguments against the **function signature** (`type_adapter.validate_python`), not against the published JSON schema. So relaxing `additionalProperties` would not have helped — and `**kwargs` is rejected by fastmcp's `from_function`. The lever that works is to **normalize arguments before validation**.

## Fix (structural — tool behavior, not guidance text)

New `_ArgTolerantFunctionTool` (overrides `run()`), applied to all three discovery tools (`search`, `tags`, `get_schema`):

- a `platform` arg is **folded into the existing `tags` filter** — platforms are catalog tags, so the model's intent is preserved and results stay relevance-scoped (verified: a `platform="central"` search returns only central-tagged tools);
- any other unknown key is dropped rather than failing the call.

The subclass is hoisted to module level (it depends only on core fastmcp, not the code-mode extra) so it is directly unit-testable.

## Verification

- Live-repro'd in the dev container: unwrapped fastmcp `Search` raises `ValidationError: platform: Unexpected keyword argument`; the tolerant tool accepts it, folds to tags, drops junk, and plain queries are unaffected.
- New behavioral tests in `tests/unit/test_server_code_mode.py` (`TestDiscoveryArgTolerance`): repro + fold + plain-query + no-tags-tool drop.
- Full suite: **1861 passed, 1 skipped**; ruff + format + mypy clean.

## Docs / version
- CHANGELOG entry; version `3.4.0.0 → 3.4.0.1` (patch). No tool-count or new-tool changes, so README/TOOLS/INSTRUCTIONS are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.4.0.1] - 2026-06-19
+
+**Patch — fix(code-mode): `search` / `tags` / `get_schema` tolerate extra args (#488).** Small local models routinely call the discovery entry point with a natural-but-undeclared argument, e.g. `search(query="sites list central", platform="central")`. fastmcp validates discovery-tool arguments against the function signature, so any unknown kwarg raised `Unexpected keyword argument` and the **entire** call was rejected — dead-ending discovery and (per external small-model testing) cascading into give-up or confabulation. This is a structural fix at the tool-behavior layer; it does not rely on guidance text.
+
+### Fixed
+- The three code-mode discovery tools (`search`, `tags`, `get_schema`) now normalize incoming arguments before validation via a new `_ArgTolerantFunctionTool` (`server.py`): a `platform` argument is **folded into the existing `tags` filter** (platforms are catalog tags, so intent is preserved and results stay relevance-scoped), and any other unknown key is dropped rather than failing the call. The published JSON schema is not fastmcp's validator, so relaxing `additionalProperties` would not have helped — the fix overrides the tool's `run()` to pre-filter arguments. (#488)
+
 ## [3.4.0.0] - 2026-06-16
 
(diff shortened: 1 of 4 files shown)


nowireless4u:
Thanks Casey — fixed in 07360c0. `_ArgTolerantFunctionTool.run()` now treats only list/tuple/set as multiple platforms; any other value (string or a malformed scalar like `123`/`true`) is coerced to a single tag, so `list()` is never called on a non-iterable. Added a regression test (`test_tolerant_search_handles_scalar_nonstring_platform`) covering `platform=123` and `platform=True`. Full suite green; ruff/format/mypy clean.

Issue #500, opened 2026-06-19
Code mode: extend uniform {items} collection shape to spec-driven mist_*/edgeconnect_* reads


nowireless4u:
## Context

#491 normalized the **Central hand-curated** collection reads to a uniform `{"items": [...]}` contract (rows always under `data["items"]`). That PR deliberately scoped out the **spec-driven** platforms:

- `mist_*` (~1000 generated tools)
- `edgeconnect_*` (~1216 generated tools)

These return whatever their upstream OpenAPI-described API returns (bare arrays, `{data: ...}`, resource-named keys, etc.), determined by the generator, not hand-written shape code. Normalizing them to `{items}` would require a generator-level transform (post-processing every generated read) — a separate, larger, higher-risk effort than the Central sweep.

## Proposal

Decide whether the uniform `data["items"]` contract should extend to the spec-driven reads, and if so, implement it at the **generator** layer (or a dispatch-time normalization in `<platform>_invoke_tool`) so the ~2200 generated tools present collections consistently.

Open questions to resolve first:
- Generator-time transform vs. runtime normalization in the invoke dispatch?
- How to detect "this response is a collection" generically across two different API styles without discarding upstream metadata.
- Whether the value justifies touching the generated surface, given these are reached almost exclusively via `<platform>_invoke_tool` (where the model already inspects the schema).

## Related
- #491 (Central hand-curated reads — done)
- Part of the code-mode small-model hardening direction (#488–#493).

PR #501, opened 2026-06-19
fix(code-mode): uniform {items} collection shape for Central reads (#491)

### LINKED ISSUE #491: Code mode: normalize inner payload shape across reads (envelope normalizes container, not `data`)
## Context

`ResponseEnvelopeMiddleware` (v3.0.0.0, #246) normalizes the **outer** envelope `{ok, status, data, message, tool, platform}` but passes `data` through **verbatim** — `src/hpe_networking_mcp/middleware/response_envelope.py:246` builds the envelope with `data=raw`, no reshaping. So while every tool reliably exposes `result["data"]`, the *shape of `data` itself* is still non-uniform across reads:

- **bare list** — `central_get_aps`
- **`{items: [...]}` dict** — the `mrt_*` reads (see #381)
- **name-keyed dict** — `central_get_site_name_id_mapping` returns `{ "<site name>": {site_id, health, ...} }` (`src/hpe_networking_mcp/platforms/central/tools/sites.py:131`)

## Symptom

A small model called `central_get_site_name_id_mapping` — a **working** call — assumed `data` was a list, so `isinstance(data, list)` was false and it reported **"0 sites."** That false-empty destroyed its confidence and it spiraled. Same class of trap as #381.

## Proposed fix (structural)

Define and enforce a uniform shape contract for collection-returning reads — e.g. always a JSON array under `data`, or always `data.items` as a list — and bring the outliers (the name-keyed dict; the bare-list vs `{items}` split) into line. The envelope cannot do this generically; it has to be a read-surface sweep, comparable in scope to the envelope rollout itself, not a one-tool tweak.

Normalizing the actual data shape (rather than documenting it in prose, which lives in untrusted channels) is the only model-agnostic fix.

---
Surfaced during external small-model testing in LM Studio (glm-4.7-flash; also reproduced on Qwen3 4B Instruct). Part of the code-mode small-model hardening set. Related: #381.

nowireless4u:
## Summary

Closes #491 — fourth of the code-mode small-model hardening set (#488–#493).

Central's hand-curated collection reads returned a **mix of shapes**: a bare list (`central_get_aps`), a name-keyed dict (`central_get_site_name_id_mapping`), and a `"No X found"` string when empty. A small model that assumed one shape silently mis-read another — the reported bug was a model treating the name-keyed mapping as a list and concluding "0 sites". All such reads now expose **one contract: rows live under `data["items"]`**, even when empty (`{"items": []}`).

## Why `{items}` (not bare list)

The original plan was a bare list, but an inventory found that most collection reads (the ~55 `mrt_*` tools, plus alerts/events) carry real pagination metadata (`total`/`next`). Flattening to a bare list would have **discarded** it. `{items}` preserves that metadata, has the smaller blast radius (most reads already use it), and gives a single rule: read `data["items"]`.

## Changed (Central hand-curated reads)

- `central_get_aps`, `central_get_ap_wlans`, `central_get_clients`, `central_get_devices`, `central_get_wlans`, `central_get_site_health` → `{"items": [...]}` (empty is `{"items": []}`, no more "No X found" string).
- `central_get_site_name_id_mapping` → `{"items": [ {site_name, site_id, health, …} ]}` (sorted worst-health first); the site name moves into each item as `site_name` (was a dict keyed by name — the actual #491 trap).
- New `as_collection()` helper (`central/utils.py`) preserving optional `total`/`next` siblings.
- `mrt_*` already return `{items}` (raw passthrough) — unchanged.

## Scope (explicit)

- **In:** Central hand-curated reads.
- **Out:** spec-driven `mist_*` / `edgeconnect_*` reads (shapes mirror upstream APIs; generator-level effort tracked in **#500**). Time-series/trend reads (`central_get_wlan_stats`) return sample arrays, not entity collections — unchanged.

## Consumers updated (blast radius from a full inventory)

- `INSTRUCTIONS.md` collection-shape guidance; `docs/TOOLS.md` entries.
- Skills (code blocks + shape tables, traced by hand since skills have no executable test): `central-site-dashboard`, `cross-platform-rf-check`, `central-ucc-quality`.
- `central_get_alerts` docstring site_id guidance.
- Unit (`test_central_monitoring`) + live integration (`test_ap_monitoring_live`, `test_wlans_live`) tests updated to the new contract.

## Verification
- Full suite: **1877 passed, 1 skipped**; ruff + format + mypy clean.

## Docs / version
- CHANGELOG entry; version `3.4.0.4 → 3.4.1.0` (**minor** — output-contract behavior change).


nowireless4u:
Good catch, fixed in the latest commit. `central_get_wlans`'s dict path now passes the upstream non-`items` siblings through: `as_collection(resp.get("items", []), **{k: v for k, v in resp.items() if k != "items"})`, so `total`/`next`/`count` survive — matching the helper's documented contract. (The other six converted reads take a bare list from a pagination-exhausting helper, so they have no siblings to lose.) Added `as_collection` unit tests covering sibling preservation + None-drop.

PR #502, opened 2026-06-19
fix(docs): scrub parrot-able example values from shipped artifacts (#492)

### LINKED ISSUE #492: Small models confabulate site data under discovery failure; audit shipped artifacts for parrot-able placeholder IDs
## Symptom

When a small model exhausted the (currently broken) discovery path, it **fabricated** a result rather than failing: it presented a site named "Headquarters – HQ" with ID `5d4e6f7a8b9c0d1e2f3a4b5c6d7e8f9a` — a perfect sequential-nibble pattern, whereas real Central site IDs are short numeric strings — as if it were live tenant data, then collapsed into repeated control-token output. For an operations tool, confabulated infrastructure data is the most dangerous failure mode of all.

## Two parts

1. **Largely mitigated by fixing discovery** (companion issues: `search` arg rejection, `Unknown tool` dead-ends, `skills_list` bloat, payload-shape normalization). A model that can actually retrieve data has no need to invent it.
2. **Docs hygiene:** audit shipped runtime artifacts — `INSTRUCTIONS.md`, `skills/*.md`, tool descriptions/docstrings — for example site/device IDs and "HQ"-style names that look real enough to be regurgitated verbatim as an answer. Prefer obviously-synthetic placeholders so a parroted example is self-evidently not live data.

Both parts are structural (fix discovery + remove parrot-able real-looking examples); neither depends on guidance text.

---
Surfaced during external small-model testing in LM Studio (glm-4.7-flash). Part of the code-mode small-model hardening set.

nowireless4u:
## Summary

Closes #492 — fifth of the code-mode small-model hardening set (#488–#493).

Part 2 of #492 (docs hygiene). Part 1 — the confabulation **root cause** — was already addressed structurally by the discovery fixes (#488–#491): a model that can retrieve data has no need to invent it. This pass removes example values realistic enough to be **regurgitated as live data**, replacing them with self-evidently synthetic placeholders.

## Audit result

A sweep of shipped runtime artifacts (`INSTRUCTIONS.md`, `skills/*.md`, tool docstrings) found **no** parrot-able hex/numeric IDs (the hallucinated `5d4e6f7a…`-style site ID has no analog in the shipped text). Only a few realistic names/addresses remained:

- `central_get_site_health` docstring — example site name `"Owls Nest"` → `"HQ"` / `"BRANCH-1"`.
- `central-qos-policy` skill — example destination subnet `165.130.0.0` (real public space; the skill is "live-verified" so plausibly real capture data) → `203.0.113.0` (RFC 5737 documentation range). The dotted-mask teaching point is preserved.
- Example emails `@corp.com` → `@example.com` (RFC 2606 reserved) in `INSTRUCTIONS.md` and `morning-coffee-report`.

## Scope

- **In:** hand-authored shipped artifacts.
- **Out:** spec-derived examples in **generated** `mist_*` / `edgeconnect_*` docstrings (a sample MAC / email from the OpenAPI spec) — regenerated from upstream, would be overwritten (related: #500). AOS firmware version strings (e.g. `8.12.0.1`) are real, meaningful values — not addresses — and are unchanged (a regex caught them; context check confirmed).

## Verification
- ruff check + `ruff format --check .` + mypy clean; targeted tests pass. No logic changed (docstring/skill/INSTRUCTIONS text + example values only).

## Docs / version
- CHANGELOG entry; version `3.4.1.0 → 3.4.1.1` (patch — shipped runtime-artifact text change).


nowireless4u:
Good catch, fixed. Switched both occurrences to `203.0.113.0/255.255.255.0` so the example sits entirely within the RFC 5737 `/24` reserved block (host bits clear, no implied real `/16`). The dotted-mask vs CIDR teaching point still lands, and it now contrasts cleanly with the nearby host example (`/255.255.255.255`).

Issue #279, opened 2026-05-07
Translations engine: data format, loader, runtime engine (foundation for cross-platform config translation)


nowireless4u:
## Background

The aos-migration skill currently emits a translation plan as text — Stage 7 hierarchy mapping + Stage 8 per-object disposition matrix + Stage 9 ordered Central API call sequence. Issue #240 tracks the eventual Phase 3 work to actually EXECUTE that plan against Central via `central_manage_*` tools.

But the underlying capability — translating a configuration from one platform's representation into another platform's API calls — has value far beyond migrations. The Mist↔Central WLAN sync use case is exactly the same primitive in non-migration form. The cross-platform `manage_wlan_profile` tool exists today because no general translation primitive existed.

This issue tracks that **general translation primitive** — three pieces:

1. **Per-target translation data files** — REST-keyed JSON describing how a target-platform configuration object is constructed from one or more source-platform objects
2. **A loader** that reads those files at startup, validates them, and exposes them via lifespan context
3. **A runtime engine** that consumes a translation plus source data plus runtime context, emits ordered API calls in the target platform's shape, with template substitution, iteration, and dependency tracking

Phase 3 (#240) is the first consumer. WLAN sync (replacing the cross-platform `manage_wlan_profile` aggregator) is the second.

### Why "translations" instead of "migrations"

- **Migration is one-time, end-state-oriented.** AOS 8 → Central is a migration. After it's done, the operator never runs it again.
- **WLAN sync is ongoing, not one-time.** Mist ↔ Central is operational reconciliation that happens repeatedly.
- **The engine doesn't care which mode the consumer is in** — it just translates source-shape to target-API-calls.

### Why "to <platform>" framing instead of pinned source-target pairs

A directional framing pinned to AOS 8 → Central becomes obsolete the moment AOS 8 migrations are complete. By keying on `targets/<platform>/<target_id>_v1.json` with platform-specific `sources` blocks, the codebase stays useful beyond AOS 8 — for example WLAN sync, Site Collection cloning, future operational automation tasks.

**v1 ships AOS 8 as the only source platform and Central as the only target platform.** Other source/target combinations are out of scope for v1; the format just doesn't preclude them.

## Empirical findings (verified against the maintainer's live Central tenant on 2026-05-07)

[...]

PR #506, opened 2026-06-23
feat(translations): canonical reader/writer engine — Mist & AOS8 → Central (forward direction)

nowireless4u:
## Summary

Foundation of the rebuilt translation engine — a **hub-and-spoke canonical model** that replaces pairwise mappers for new source/target pairs. References #279.

- **Readers** turn a platform's config into a platform-neutral `CanonicalWlan`.
- **Writers** turn canonical into ordered target API-call descriptors.
- An **orchestrator** plans + executes them with ensure-or-create idempotency.

This ships the **forward** direction end-to-end — **Mist→Central** and **AOS8→Central** WLAN + RADIUS — live-verified against the tenants.

## Scope / what this is (and isn't)

- **Additive and not yet wired into any tool or skill.** The existing JSON `emit_calls` engine still backs the AOS8-JSON translations. Nothing in production calls the new path yet.
- The **cutover** (repoint `central_translation_preview` / tooling onto the new engine, retire the JSON path for non-AOS8) is a **follow-up PR**.
- The **reverse direction** (Central→Mist, AOS8→Mist) is a **follow-up PR**.
- No user-facing tool changes → tool counts and version-gated docs are unchanged. Version bumped (3.4.2.0) + CHANGELOG only, since shipped `src/` artifacts changed.

## What's included

| Module | Role |
|---|---|
| `canonical/wlan.py` | neutral WLAN model — granular `(key_mgmt, wpa_version, cipher)` opmode triplet, VLAN, rates, perf, isolation, WMM, forward-mode (bridged/tunneled/hybrid + dual + clusters), assignment facets |
| `readers/mist.py` | Mist WLAN → canonical (security, RADIUS auth/acct/CoA incl. `{{var}}`, VLAN, template assignment) |
| `readers/aos8.py` | AOS8 (source-only) → canonical: `virtual_ap`↔`ssid_prof`↔`aaa_prof` join, `aaa_prof → server_group_prof → rad_server` RADIUS chain, opmode→triplet, target-mode→forward/dual |
| `writers/central.py` | canonical → `wlan-ssids` (library create + GLOBAL/SITE/SITE_COLLECTION/DEVICE_COLLECTION assignment), tunneled/hybrid `overlay-wlan` cluster binding, dual-mode |
| `writers/central_radius.py` | canonical RADIUS → `{ssid}_nac` server-group + auth-servers (consolidated by host), `{{var}}`→`ALIAS_AUTH_SERVER_ADDRESS` |
| `orchestrator.py` | registry, pure `plan()`, async `execute()` (ensure-or-create), Central scope resolution |

## Testing

- **99 unit tests** across canonical/readers/writers/orchestrator; full suite **1981 passed**; ruff/format/mypy clean.
- **Live-verified** end-to-end:
  - Mist `ZZRADTEST` (literal + `{{var}}` RADIUS hosts) → Central: alias + auth-servers + server-group + wlan-ssids, idempotent re-run skips, cleanup clean.
  - AOS8 `/md/Campus/West` real WLANs (OWE/WPA2-DOT1X/WPA3-SAE/MPSK) → correct opmode/triplet; full AOS8→Central round-trip via the orchestrator.
  - Central tenant is monitor-only (no managed devices) — safe to fire; left clean.

## Notes for review

- Two latent bugs fixed in passing: numeric VLAN now emits `VLAN_RANGES`/`vlan-id-range` (was `NAMED_VLAN`); `SAE` maps to `WPA3_SAE` (the prior `WPA3_PERSONAL` is not a valid Central opmode).
- AOS8 server-groups key on `sg_name` (not `profile-name`) — found + fixed via live testing.


nowireless4u:
Thanks Casey — both execution catches were spot on. Addressed in f2557bf:

**1. Unresolved gateway clusters now block.** `TranslationPlan.unresolved` now also surfaces calls flagged `unresolved_clusters` (as `{"kind": "gateway_cluster", ...}`), so `execute()`'s existing pre-write block covers tunneled/hybrid/dual WLANs with no cluster binding — same treatment as unresolved assignment scopes. Regression test: `test_execute_blocks_on_unresolved_clusters`.

**2. `execute()` now honors `depends_on`.** It tracks completed indices and skips any call whose prerequisites didn't complete, returning `blocked_dependency_failed` (existing/dry-run count as complete). A failed alias/auth-server/server-group create now blocks the dependent WLAN/assignment instead of running against a half-built config. Regression test: `test_execute_blocks_dependents_on_failure`.

**Security green:**
- Bandit B105 — the `{"secret": None}` placeholder is marked `# nosec B105` (the real RADIUS secret is filled from source below it).
- pip-audit — `uv.lock` bumped `msgpack 1.1.2 → 1.2.1` and `pydantic-settings 2.14.0 → 2.14.2`.

Full suite 1983 passed; bandit/ruff/mypy clean locally.

nowireless4u:
Good catch again — wired in f-up commit.

`central_write_wlan()` now points the assignment `depends_on` at the **overlay call index** when `canon.forward` is tunneled/hybrid (bridged still depends on the WLAN create only). With the executor's dependency guard already in place, a failed `overlay-wlan` POST now blocks the scope assignment instead of activating an unbound tunneled WLAN.

Tests added:
- `test_tunneled_assignment_depends_on_overlay` / `test_bridged_assignment_depends_on_create_only` (writer call graph)
- `test_execute_overlay_failure_blocks_assignment` (end-to-end: overlay POST fails → assignment returns `blocked_dependency_failed`, nothing assigned)

Full suite 1986 passed; bandit/ruff/mypy clean.

nowireless4u:
Fixed in follow-up — and yep, our live `MIST-NAC` WLAN would have hit exactly this.

- `_wlan_body()` now gates the `{ssid}_nac` `auth-server-group` / `acct-server-group` reference on `auth_source.kind == RADIUS_GROUP`. A NAC auth source gets no reference (nothing creates that group).
- A NAC-backed WLAN's create is flagged `unresolved {"kind": "nac_profile"}`, so `plan.unresolved` is non-empty and `execute()` returns `blocked_unresolved` — it won't create an enterprise WLAN with no valid auth backend. The Central NAC reference will be emitted once the NAC canonical+writer lands (a planned follow-up).

Tests: `test_nac_wlan_no_server_group_ref_and_flagged_unresolved`, `test_radius_group_wlan_is_not_flagged_unresolved`. Full suite 1988 passed; security/lint/type clean.

PR #508, opened 2026-06-23
feat(translations): reverse direction — Central reader + Mist writer (Central→Mist, AOS8→Mist)

nowireless4u:
## Summary

Adds the **reverse** of the [#506](https://github.com/nowireless4u/hpe-networking-mcp/pull/506) forward foundation: a **Central reader** + **Mist writer** so a Central WLAN (or an AOS8 WLAN) can be translated *to* Mist through the same canonical model + orchestrator. References #279.

## Scope

- **Additive / not wired into any tool yet.** The cutover that repoints tooling onto the engine is a later PR.
- No user-facing tool changes → tool counts / version-gated docs unchanged. Version bumped 3.4.3.0 + CHANGELOG (shipped `src/` artifacts changed).
- Reverse direction completes the bidirectional WLAN story: **Mist↔Central** + **AOS8→{Central,Mist}** (AOS8 stays source-only).

## What's included

| Module | Role |
|---|---|
| `readers/central.py` | Central `wlan-ssids` → canonical (inverts the Central writer): opmode→triplet, essid literal/alias, VLAN, perf/isolation/WMM; **RADIUS walk** `auth-server-group → server-group → auth-servers` (incl. AUTH_AND_COA) → auth/acct/CoA; `ALIAS_AUTH_SERVER_ADDRESS` → `{{var}}`; `sys_central_nac` → NAC; assignment from `config-assignment` rows |
| `writers/mist.py` | canonical → Mist per-WLAN **template** (`applies` by name→id, unresolved flagged) + org WLAN; triplet → `auth.type`+`pairwise`; **inline** `auth_servers`/`acct_servers`/`coa_servers`; NAC → `mist_nac`; cloud-MPSK → `dynamic_psk`; `{{var}}` passthrough |
| `orchestrator.py` | registered `central:wlan` reader + `mist:wlan` writer; `execute()` gained general **capture/inject** (Mist is ID-based — the template create captures its `id`, the WLAN injects it as `template_id`) + an `idempotent` flag so Mist collection-POSTs skip the Central GET-exists probe |

## Testing

- **34 new unit tests** (Central reader, Mist writer, executor capture/inject + idempotent). Full suite **2022 passed**; bandit/ruff/mypy clean.
- **Live-verified** end-to-end: seeded a Central enterprise WLAN (+ server-group) → read it back via the Central reader → canonical → Mist writer → **created in Mist** (template + WLAN, inline RADIUS, `auth eap`, `vlan_id` correct, `template_id` threaded via capture/inject) → cleaned up **both** sides. Central is monitor-only/fire-safe; the Mist template was created with empty `applies` (never broadcasts to live sites).

## Notes for review

- NAC detection on the Central side keys on the system `sys_central_nac` group (heuristic); the full Central-NAC ↔ Mist-NAC object mapping rides with the planned NAC canonical+writer.
- `{{var}}` reverse: an `auth-server-address` that isn't an IP/FQDN literal (or matches a defined alias) maps back to `{{var}}` — Mist resolves vars per site.


nowireless4u:
Good catch — fixed in 9842a2e.

`TranslationPlan.unresolved` now flattens both shapes: a call's `unresolved` may be a single dict (Central writer, one scope) or a list of dicts (Mist template, several scopes from one call) — it `extend`s lists and `append`s dicts into one flat list of `{kind, name}`. So `preview()` no longer trips on the Mist case, and a writer can report multiple unresolved scopes from one call.

Regression test `test_mist_unresolved_scope_flattens_and_blocks`: a Mist plan with an unresolved site → `plan.unresolved` is a flat dict list, `preview()` runs, and `execute()` returns `blocked_unresolved` with nothing written. Full suite 2023 passed.

nowireless4u:
Fixed in 73788a8 (+ merged latest main).

WEP now blocks instead of emitting empty auth: `_auth()` had no `WEP_STATIC`/`WEP_DYNAMIC` branch, so those fell through to `auth: {}`. WEP has no clean Mist mapping in this writer, so the WLAN call is now flagged `unresolved {"kind": "unsupported_auth"}` and the plan blocks before any write — same contract as NAC and unresolved scopes (rather than a silent create with unintended auth defaults).

Tests: `test_wep_is_flagged_unsupported` (writer flags both WEP modes) and `test_aos8_wep_to_mist_blocks` (AOS8 `static-wep` → Mist plan → `execute()` returns `blocked_unresolved`, nothing written). Full suite 2026 passed.

PR #509, opened 2026-06-24
feat(translations): cutover — translate_wlan tools + retire the old WLAN-sync surface

nowireless4u:
## Summary

The canonical translation engine (#506 forward, #508 reverse) was fully built but **unwired** — nothing called it. This **cutover** exposes it through two cross-platform bridge tools and retires the legacy hand-mapping path. References #279. (PR3 of the rebuild.)

## Added

- **`translate_wlan_preview`** (read-only) — given `(source_platform, target_platform, ssid)`, fetches the source WLAN + *all* engine context server-side (reusing the engine's resolvers) and returns the ordered target calls for review. **Secrets (PSK / RADIUS / CoA) are redacted and call bodies omitted** from preview output.
- **`translate_wlan_apply`** — executes the translation against the target via `orchestrator.execute`. **Gated**: it checks the *target* platform's `ENABLE_*_WRITE_TOOLS` flag at runtime (the target varies per call, so a static write-tag would hide the tool for the wrong flag) and fires the universal elicitation confirmation before writing.
- Both are registered in **every** mode (incl. code) — reachable from `execute()` via `call_tool(...)`, like `health` — because the engine can't be imported inside the sandbox.
- `tests/unit/test_translate_wlan.py`.

Sources: `mist` / `central` / `aos8` (AOS8 via `source_override`). Targets: `central` / `mist`. Central command adapter = `central_conn.command`; Mist adapter wraps the raw httpx client (not `mist_request`, which raises on 4xx and would abort `execute()`).

## Removed (breaking)

`manage_wlan_profile`, the `sync_wlans_*` prompts, and the internal `wlan_mapper` / `_wlan_helpers` modules. These were dynamic-mode-only and returned hand-mapping *instructions* for the AI — exactly what the engine replaces. Scripts using `manage_wlan_profile` should switch to `translate_wlan_preview` / `translate_wlan_apply`. **`central_manage_wlan_profile` (per-platform write tool) is unaffected.** `aos-migration.md` (uses the old JSON engine via `central_translation_preview`) is unaffected.

## Changed

`wlan-sync-validation` skill repointed (drift correction → preview then apply). Docs updated: README, docs/TOOLS.md, INSTRUCTIONS.md, CHANGELOG. Version → 3.4.4.0.

## Testing

- Full suite **1953 passed** (the 81 `wlan_mapper` tests removed with the module; 7 new translate_wlan tests); ruff / format / mypy / bandit clean.
- **Live-verified** (dev container, real tenants): preview `mist→central ZZRADTEST` redacts `secret-*`, omits bodies, lists the 5 ordered calls; apply creates then re-run is fully `skipped_exists` (idempotent); `ENABLE_CENTRAL_WRITE_TOOLS=false` → `403`; cleanup `200`.

## Notes for review

- Mist `apply` is **not** idempotent (collection POSTs → re-run duplicates); documented in the tool + skill. Central is idempotent (ensure-or-create).
- AOS8-by-SSID isn't a clean primitive (SSID lives in `ssid_prof`); v1 requires `source_override`/`context_override` for AOS8 source.
- Out of scope (follow-ups): NAC canonical+writer; repointing `aos-migration.md`'s WLAN stage off the old JSON engine.


nowireless4u:
Good catch — the param was wired to the reader but never reached the Central writer. Fixed in the latest commit.

Rather than guess the cluster entries (cluster-type / scope-id / tunnel-type are topology decisions, and there are no gateway clusters in the tenant to derive or live-verify against), I exposed a **first-class public param** `gateway_cluster_list` — the resolved Central overlay entries — on both tools. `_build_plan` merges it into `writer_ctx` for central targets (explicit param wins over `context_override`), so callers no longer touch the internal wiring. `gateway_clusters` (names) stays as the AOS8 reader hint. This matches how the existing engine sources cluster bindings (operator/skill supplied).

Regression tests added per your list (AOS8 source, `target_mode="tunneled"`, non-empty `gateway_cluster_list`):
- `test_tunneled_with_cluster_list_resolves_overlay` — the overlay call's `gw-cluster-list` is non-empty **and** `plan.unresolved` excludes the gateway cluster.
- `test_tunneled_without_cluster_list_blocks` — without it, the cluster is `unresolved` (apply blocks).

Full suite 1955 passed; docs updated. (The overlay-wlan create itself was live-verified in #506; this change only threads the resolved list through the bridge.)

Issue #532, opened 2026-06-25
Code mode: response-envelope / collection-shape navigation → 'str' object has no attribute 'get', no self-correcting hint


nowireless4u:
**Reported by Zach (Claude Haiku 4.5), two separate sessions, same failure. Found alongside the code-mode small-model sandbox audit (follow-up to #512) — but this is a DISTINCT class from #513–#519 (those are sandbox language/stdlib limits; this is response-shape navigation).**

### Symptom
```python
resp = await call_tool("mist_list_tools", {"filter": "stat"})
names = [tool.get("name") for tool in resp]      # <-- iterates the dict
# -> AttributeError: 'str' object has no attribute 'get'
```
Second report identical with `{"filter": "count.*client"}` and `for tool in tools_resp: tool.get("name")`.

### Root cause
Two stacked shape surprises, both leading to the same `'X' object has no attribute 'get'` error (the #1 code-mode AttributeError trap):

1. **Everything is enveloped.** `call_tool` results are wrapped by `ResponseEnvelopeMiddleware`: `{"ok", "status", "data", "message", "tool", "platform"}`. `mist_list_tools` is NOT in the envelope bypass list, so its payload is under `result["data"]`.
2. **`list_tools` nests its list.** Even unwrapped, `mist_list_tools` returns `{"platform", "total", "categories", "tools": [...]}` — the tool dicts are under `["tools"]`, not a bare list.

So iterating the response yields **string keys** (`"ok"`, `"status"`, …), and `str.get` doesn't exist. Correct access: `resp["data"]["tools"]`, then each element has `.get("name")` / `.get("summary")`.

A small model has no way to discover this from the error alone, and INSTRUCTIONS.md guidance is advisory (treated as untrusted) — so the fix must be structural.

### Proposed fix
Add a `SandboxErrorCatchMiddleware` rule for the shape-mismatch family — `'<type>' object has no attribute 'get'` (also `'keys'` / `'items'`) — appending a self-correcting hint:

> You're calling a dict method on a non-dict (often a string from iterating a dict's keys). `call_tool` results are wrapped in an envelope — read the payload via `result["data"]`. Collection shape varies: `<platform>_list_tools` → `result["data"]["tools"]`; many monitoring reads → `result["data"]["items"]`; some reads are a bare list under `result["data"]`. Check the shape (e.g. `isinstance(x, dict)`) before `.get()`.

### Related
[...]

Issue #519, opened 2026-06-25
Code mode: unsupported syntax (with / match / class / del) — consider substitute hints


nowireless4u:
**Found in the code-mode small-model sandbox audit (follow-up to #512).**

The monty parser rejects several syntax constructs a model may emit. Unlike the runtime traps (modules, `.format`, `next`), these already fail with a *descriptive* `NotImplementedError` ("the monty syntax parser does not yet support …"), so they are partially self-correcting already — hence lower priority.

### Verified unsupported (SyntaxError / NotImplementedError at parse time)
- `with` statements (context managers)
- `match` / `case` (structural pattern matching)
- `class` definitions
- `del` statements
- `yield` / `yield from` (already documented)

### Already supported
`async`-free nested `def`, decorators, `try/except/else/finally`, `assert`, `global`, `lambda`, comprehensions, walrus `:=`, unpacking, `*args/**kwargs`.

### Proposed fix (optional / P3)
The descriptive parse error is usually enough. If we want belt-and-suspenders, add brief substitute hints (e.g. `with` → manual acquire/use; `class` → use a plain `dict`; `del d[k]` → `d.pop(k, None)`; `match` → `if/elif`). Mostly this is an INSTRUCTIONS.md "blocked constructs" table refresh rather than middleware work.

nowireless4u:
Resolved on `main` (PR #542, commit `9313f87`); tracker left open. This was scoped in the issue itself as P3/docs-only — a blocked-constructs substitute-table refresh, not middleware work. Shipped: `src/hpe_networking_mcp/INSTRUCTIONS.md:280-283` adds substitute rows for `with`, `match`/`case`, `class`, and `del` (the sandbox parser already emits descriptive `NotImplementedError`s for these, so no hint function was requested). Closing as its intended docs-only scope.

Issue #518, opened 2026-06-25
Code mode: dict | dict merge unsupported — raw TypeError, no {**a, **b} hint


nowireless4u:
**Found in the code-mode small-model sandbox audit (follow-up to #512).**

The PEP 584 dict-union operator (`a | b`) is not supported in the monty sandbox. The older spread form works.

### Repro
```python
return {'x': 1} | {'y': 2}
# -> TypeError: unsupported operand type(s) for |: 'dict' and 'dict'
```

### Works
```python
return {**{'x': 1}, **{'y': 2}}   # ok
```

### Proposed fix
Add a `SandboxErrorCatchMiddleware` rule: when the cause matches `unsupported operand type(s) for |: 'dict' and 'dict'`, append a hint — "the sandbox doesn't support `dict | dict`; use `{**a, **b}` to merge." Note in INSTRUCTIONS.md. (Lower priority than the module / `.format` / `next` traps.)

nowireless4u:
Resolved on `main` (PR #537, commit `649c4d6`); tracker left open. `dict | dict` unsupported-operand now returns a `{**a, **b}` merge hint — `sandbox_error_catch.py:238-242` (`_dict_union_hint`). Test: `test_middleware.py` case `{"a":1} | {"b":2}`. Verified green.

Issue #517, opened 2026-06-25
Code mode: next(generator, default) unsupported — raw TypeError, no loop hint


nowireless4u:
**Found in the code-mode small-model sandbox audit (follow-up to #512).**

`next(<generator-expression>, <default>)` fails in the monty sandbox with a misleading `TypeError`. This is a recurring trap — it previously bit a shipped skill snippet and had to be rewritten as an explicit loop.

### Repro
```python
return next((x for x in [] if x > 1), 'd')
# -> TypeError: 'list' object is not an iterator
```

### Works
```python
g = iter([])
return next(g, 'd')   # ok on an explicit iterator
```

### Proposed fix
Add a `SandboxErrorCatchMiddleware` rule: when the cause matches `'...' object is not an iterator` (the `next(genexpr, default)` shape), append a hint — "the sandbox can't use `next(<generator>, default)`; use an explicit loop, a comprehension + index (`matches = [x for x in xs if cond]; matches[0] if matches else default`), or `next(iter(seq), default)`." Reinforce in INSTRUCTIONS.md (already mentioned for `next()` historically; make it a runtime hint).

nowireless4u:
Resolved on `main` (PR #537, commit `649c4d6`); tracker left open. `next(<generator>, default)` "not an iterator" now returns a comprehension / `next(iter(...))` hint — `sandbox_error_catch.py:227-235` (`_next_iter_hint`). Test: `test_middleware.py` case `next((x for x in [] if x>1),"d")`. Verified green.

PR #537, opened 2026-06-25
fix(code-mode): self-correcting sandbox hints + #513 regression fix (#513,#514,#515,#516,#517,#518,#532)

### LINKED ISSUE #513: Code mode: NameError self-correct hint misfires on unavailable builtins (regression in #512)
**Found in the code-mode small-model sandbox audit (follow-up to #512).**

#512 added a `SandboxErrorCatchMiddleware` hint that turns `NameError: name 'X' is not defined` into a "each `execute()` is stateless — recompute `X` / keep work in one block" message. That's correct for the cross-block-state case (the Haiku `org_id` failure), but it **misfires** when `X` is an *unavailable builtin* rather than an operator variable.

### Repro
```python
return getattr(object(), 'x', 1)
# -> NameError: name 'object' is not defined
```
`object` (and `open`, `eval`, `exec`, `input`, `compile`, `globals`, `locals`, `vars`, …) are not bound in the monty sandbox. They produce the **same** `NameError` shape as a cross-block variable, so the current hint wrongly tells the model to "recompute `object` from a previous block."

### Proposed fix
In the NameError hint, distinguish the two causes:
- If the undefined name is a known Python builtin that the sandbox doesn't provide → say "`X` is a builtin that isn't available in the sandbox" (+ substitute where one exists).
- Otherwise → keep the existing cross-block-statelessness / typo message.

A small curated set of known-unavailable builtins gets the crisp message; everything else can fall back to a multi-cause message ("from a previous block (state doesn't persist), an unsupported builtin, or a typo").

### Verification
Probed 76 common constructs against pydantic-monty with the production `os=OSAccess()` config — `object` is one of the names that returns this NameError.

nowireless4u:
Group C/D of the small-model backlog — the sandbox self-correcting hints from the audit. Closes #513, #514, #515, #516, #517, #518, #532.

`SandboxErrorCatchMiddleware`'s hint logic is now an **ordered rule table** (first match wins). Each common raw sandbox error becomes an actionable hint in the error result — the one channel small/local models reliably act on (INSTRUCTIONS.md is advisory). All hints are driven/tested against **real monty errors**, not synthetic strings.

| # | Dead-end | Hint |
|---|---|---|
| **#513** (regression) | `NameError` on a missing builtin (`object`/`open`/…) | "unavailable builtin" message instead of the misleading "recompute from a previous block" advice. Genuine undefined vars still get the statelessness hint. |
| **#514** | `ModuleNotFoundError` | most stdlib absent + builtin substitutes (`collections`→dict, `statistics.mean`→`sum/len`, `itertools`/`functools`→loops) |
| **#515** | `str.format()` | use f-strings |
| **#516** | `json.dumps(default=…)` | drop `default=`, pre-convert |
| **#517** | `next(<generator>, default)` | comprehension+index / `next(iter(seq), default)` |
| **#518** | `dict \| dict` | `{**a, **b}` |
| **#532** | `'<type>' object has no attribute 'get'` | envelope (`result["data"]`) + collection shape (`["tools"]`/`["items"]`/bare list) — the #1 shape trap, reported by Zach twice |

### #513 detail (the regression I shipped in v3.4.5.1)
Missing builtins raise the *same* `NameError` shape as a cross-block-state slip, so the v3.4.5.1 hint told the model to "recompute `object` from a previous block." Now a curated set of unavailable builtins gets a distinct message; everything else keeps the statelessness lead (with builtin/typo mentioned).

### Also
- INSTRUCTIONS.md "known-blocked" table expanded to match (stdlib modules, `.format()`, `json default=`, `next(gen)`, `dict |`, missing builtins, envelope/shape navigation).
- Tests: parametrized real-error → hint coverage for all seven, plus a #513 regression guard (`object`→builtin msg, `org_id`→statelessness msg).
- Full Docker suite green: ruff + format + mypy + 1748 passed / 1 skipped.

Patch bump 3.4.5.3 → 3.4.5.4.

---
Remaining: #522+#533 (envelope consistency), #524–#527 (discovery), #519/#528/#529/#531 (docs). **#530 obsolete** (files deleted in #510).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.4.5.4] - 2026-06-25
 
-**Patch — fix(code-mode): self-correcting hints for the common small-model sandbox dead-ends + fix a #512 regression (#513, #514, #515, #516, #517, #518, #532).** `SandboxErrorCatchMiddleware`'s hint logic is now an ordered rule table; each common raw sandbox error is turned into an actionable, self-correcting message in the error result (the channel models reliably act on).
+**Patch — fix(code-mode): self-correcting hints for the common small-model sandbox dead-ends + fix the #513 regression (introduced in v3.4.5.1 / #512) (#513, #514, #515, #516, #517, #518, #532).** `SandboxErrorCatchMiddleware`'s hint logic is now an ordered rule table; each common raw sandbox error is turned into an actionable, self-correcting message in the error result (the channel models reliably act on).
 
 ### Fixed
 - **#513 (regression in v3.4.5.1)** — the `NameError` hint no longer misfires on unavailable builtins. `object` / `open` / `eval` / … now get an "unavailable builtin" message instead of the misleading "recompute from a previous block" advice; genuine cross-block variables still get the statelessness hint.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.4.5.4] - 2026-06-25
+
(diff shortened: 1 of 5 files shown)


nowireless4u:
Fixed — CHANGELOG now reads "fix the #513 regression (introduced in v3.4.5.1 / #512)", matching the body and the issue list. Thanks for the catch.

PR #540, opened 2026-06-26
chore(platforms): future-proof tool registration against the #524 ordering trap

nowireless4u:
Follow-up to #539 (the Mist registration fix). You asked to make sure the platform template registers tools properly so future platforms don't hit the same bug — here's the audit result and the guard.

## Audit (Mist vs Central vs everyone)
The #524 bug needs **two** things together: (a) `tools/__init__` eager-imports its submodules, **and** (b) `register_tools` imports the tools package *before* setting `_registry.mcp = mcp`. Only **Mist** had both.

| platform | `tools/__init__` | mcp before tools import? | status |
|---|---|---|---|
| **template** | lazy | yes (importlib after wiring) | ✅ already safe |
| central / clearpass / apstra / axis / greenlake / aos8 / uxi | lazy | importlib-only | ✅ safe |
| **edgeconnect** | lazy | **no** (static `import tools` before wiring) | ⚠️ safe today, latent |
| mist | eager | yes (fixed in #539) | ✅ |

The **template is already correct** (wires `mcp` first, imports each category via `importlib` after, lazy `__init__`). Mist broke only because it *diverged* into the eager-package + pkgutil pattern.

## Changes
- **Template**: explicit ORDER-MATTERS comment on `_registry.mcp = mcp` so the contract is impossible to miss when copying — especially if a new platform switches to the eager/pkgutil pattern Mist used.
- **EdgeConnect**: defensive reorder — wire `_registry.mcp` before importing its `tools` package, removing the latent trap (it's safe today only because its `__init__` is lazy).
- **New `test_platform_registration_order.py`**: parametrized over all 10 platforms (+ template). Asserts (1) any static `tools`-package import comes after `_registry.mcp = mcp`, and (2) an eager `tools/__init__` requires mcp wired first. Source-level so import caching can't mask it — a future platform that reintroduces the trap fails CI.

## Tests
ruff + format + mypy + 1777 passed / 1 skipped (incl. 20 new parametrized cases).

Patch bump 3.4.5.6 → 3.4.5.7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.4.5.7] - 2026-06-26
+
+**Patch — chore(platforms): future-proof tool registration against the #524 ordering trap.** Follow-up to the Mist registration fix (#524): make sure no future platform can reintroduce the "tools recorded in the registry but never registered with FastMCP" bug.
+
+### Changed
+- **Template hardened** — `platforms/_template/__init__.py` now carries an explicit ORDER-MATTERS comment: `_registry.mcp = mcp` must be set before importing any tool module (or `@tool` records the spec but skips `mcp.tool()`). The template was already correct; this makes the contract impossible to miss when copying it.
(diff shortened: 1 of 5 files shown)


nowireless4u:
Good catch — fixed. The test now discovers platforms dynamically from `src/hpe_networking_mcp/platforms` (any subdirectory whose `__init__.py` defines `register_tools`, which naturally excludes `_common` and includes `_template`), so a newly added platform is guarded automatically without anyone remembering to update a list. Added a `test_platform_discovery_is_sane` guard-the-guard check so discovery can't silently degrade to a vacuous pass (asserts the known platforms are found, ≥9 total).

Issue #546, opened 2026-06-30
SECURITY: read_file exposes raw uploaded file content to the model (serials/MACs, AOS 8 PSKs/RADIUS secrets)


nowireless4u:
## Summary
With `MCP_APP_ENABLE=true`, the MCP-Apps `FileUpload` provider's **`read_file`** tool is model-visible and returns raw uploaded file **content** to the LLM. This defeats the purpose of server-side uploads: a model can read any uploaded file into its context.

Observed live: after uploading a device CSV for `greenlake_bulk_add_devices`, an MCP-Apps client called `read_file` to "verify the format" and pulled the device serial + MAC straight into context. The same exposure applies to `aos-migration` config uploads, which carry **PSKs and RADIUS/TACACS secrets**.

## Impact
Sensitive upload contents (device serials/MACs, AOS 8 PSKs, RADIUS/TACACS shared secrets) reach the model context — exactly what the "upload bypasses the context window" design promises to prevent.

## Secondary leak
`greenlake_bulk_add_devices`' `build_result_envelope` returns **raw serials** in its `failures` / `enrichment_failures` lists. GreenLake is not yet in the PII tokenization ruleset, so these are not scrubbed by middleware either.

## Fix
- Remove `read_file` from the model surface entirely (it is provider-registered; `remove_tool` doesn't work on provider tools, so a `Visibility(False, names={"read_file"})` transform is used — hidden AND uncallable). `file_manager` (UI) + `list_files` (metadata only) remain.
- Uploaded files are read **server-side only** via `provider.on_read` (shared `utils/uploads.read_uploaded_text`). Add `aos8_parse_config(filename=...)` so `aos-migration` no longer needs `read_file`.
- Tokenize serials in `build_result_envelope`.

## Principle
No model-visible tool returns raw uploaded file content. Consumers read by name inside the server and return only counts / tokenized / parsed results.

PR #549, opened 2026-06-30
fix(greenlake): tokenize bulk-add serials (lifespan token_store wiring + camelCase rule)

### LINKED ISSUE #548: GreenLake bulk-add serials redact to bare [serial] (token_store not on lifespan_context; camelCase serialNumber uncovered)
Follow-up surfaced by the v3.4.6.2 live upload test (#546). No raw serial/MAC reached the model (good), but:

1. **`token_store` is never placed on `lifespan_context`** — only `mcp._hpe_token_store` is set in `create_server`. So `greenlake_bulk_add_devices`' serial tokenization (`_safe_serial`) always falls back to the bare `[serial]` placeholder. Consequences: an MCP-Apps client misread `[serial]` as "the CSV has unfilled template text," and a real multi-device failure list shows identical `[serial]` for every failure (can't identify which device).
2. **GreenLake camelCase `serialNumber` is uncovered** — the field-name normalizer lowercases but doesn't split camelCase, so `serialNumber` → `serialnumber`, which wasn't a ruleset key. GreenLake device serials therefore weren't tokenized even with PII enabled.

## Fix
- Put `token_store` on `lifespan_context` so the per-session `Tokenizer` is built and serials become `[[SERIAL:uuid]]` tokens (round-trip with `ENABLE_PII_TOKENIZATION=true`; non-leaking `[serial]` default retained).
- Add `serialnumber` to `TOKENIZED_IDENTIFIER_FIELDS` for GreenLake API responses. (MACs are normalized, not tokenized — no MAC `TokenKind` by design.)

nowireless4u:
## Context
Follow-up to the v3.4.6.2 live upload test (#546). That test confirmed the security win — **no raw serial/MAC reached the model** — but the bulk-add result redacted serials to a bare `[serial]` placeholder, which:
- an MCP-Apps client **misread** as "the CSV has unfilled template text," and
- makes a real multi-device failure list useless (every failure shows identical `[serial]`).

## Root cause
`token_store` was **never placed on `lifespan_context`** — `create_server` only sets `mcp._hpe_token_store`. So `greenlake_bulk_add_devices` built no per-session `Tokenizer` and `_safe_serial` always hit the fallback. Separately, GreenLake's camelCase `serialNumber` normalizes to `serialnumber` (the field-name normalizer lowercases but doesn't split camelCase), which wasn't a ruleset key — so GreenLake serials weren't tokenized even with PII on.

## Fix
- **`token_store` on `lifespan_context`** → serials emit as `[[SERIAL:uuid]]` tokens. With `ENABLE_PII_TOKENIZATION=true` they round-trip (operator sees the real serial via detokenization; the AI's context holds only the token). The non-leaking `[serial]` default is retained for the no-store case.
- **`serialnumber` added to `TOKENIZED_IDENTIFIER_FIELDS`** so GreenLake API responses (e.g. `greenlake_get_devices`) tokenize device serials. MACs are normalized, not tokenized (no MAC `TokenKind`, by design — MACs are on-wire observable).
- The tokenizer is idempotent (`TOKEN_RE.fullmatch` guard), so `_safe_serial` pre-tokenizing + the walker re-processing the `serial` field doesn't double-tokenize.

## Validation
- `test_pii_redaction`: `serialNumber`/`serialnumber`/`serial`/`serial_number` all classify as `TOKENIZE_IDENTIFIER` / `SERIAL`.
- `test_greenlake_bulk_add`: with `token_store` on `lifespan_context`, failed-row serials are `[[SERIAL:uuid]]` tokens, raw serial absent, not the `[serial]` fallback.
- Full gate green: ruff + format + mypy (682 files) + **1959 passed / 1 skipped**.

## Scope
No tool/parameter/count changes. Round-trip + GreenLake-wide coverage activate only with `ENABLE_PII_TOKENIZATION=true`; bulk-add serial redaction itself is always on. Patch bump 3.4.6.2 → 3.4.6.3 (uv.lock re-locked).

Closes #548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.4.6.3] - 2026-06-30
+
+**Patch — GreenLake device serials become round-trippable tokens (fixes a latent tokenization gap).** The live upload test for v3.4.6.2 confirmed no raw serial/MAC reaches the model, but the bulk-add result redacted serials to a bare `[serial]` placeholder — which (a) an MCP-Apps client misread as "the CSV has unfilled template text," and (b) makes a real multi-device failure list useless (every failure shows identical `[serial]`). Root cause: `token_store` was never placed on `lifespan_context`, so `greenlake_bulk_add_devices`' serial tokenization silently no-op'd to the fallback.
+
+### Fixed
+- **`token_store` is now on `lifespan_context`** — `greenlake_bulk_add_devices` builds its per-session `Tokenizer` from it, so failed-row serials are emitted as `[[SERIAL:uuid]]` tokens instead of the bare `[serial]` placeholder. With `ENABLE_PII_TOKENIZATION=true` these round-trip (the operator sees the real serial via detokenization; the AI's context only ever holds the token). The non-leaking `[serial]` default remains as a defensive fallback.
+
+### Changed
+- **GreenLake camelCase identifiers join the PII ruleset** — added `serialnumber` (the lowercased form of GreenLake's `serialNumber`; the field-name normalizer doesn't split camelCase) so device serials in GreenLake API responses (e.g. `greenlake_get_devices`) tokenize when PII tokenization is enabled. (MACs are normalized, not tokenized — by design, there is no MAC `TokenKind`.)
+
+### Notes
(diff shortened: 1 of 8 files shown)


nowireless4u:
Good catch — fixed. `_safe_serial` now catches `KeymapFullError` and returns the non-leaking `[serial]` fallback, so a full session keymap can't crash the write tool after it's already done its POST/assignment work. Added a regression test (`test_full_keymap_does_not_crash_write_tool`) that runs bulk-add with `TokenStore(max_entries_per_session=0)` and asserts it completes with `[serial]` rather than raising. Full gate green (1960 passed / 1 skipped).

PR #550, opened 2026-06-30
feat(greenlake): device-onboarding runbook + batch-uniform assignment params

nowireless4u:
## Why
From an operator transcript: prompting "I want to add devices to GreenLake" had the AI call `greenlake_bulk_add_devices` bare, which (1) never warned that pasted data is visible to the AI, (2) didn't really prompt for upload-vs-paste — it just showed the widget and offered paste, and (3) only *added* devices. Real onboarding is a lifecycle: **add → subscribe → assign to a service**. Nobody adds devices and comes back later to subscribe them.

## What
- **New `greenlake-device-onboarding` skill** — the guided front-end (18 bundled skills now). Stages: get the list (**choose upload vs paste, with the mandatory paste-visibility warning**) → pick a subscription (`greenlake_get_subscriptions`) → assign a service → optional location/tags → run → verify. The AI is told NOT to call the bare tool for onboarding requests; INSTRUCTIONS.md maps the trigger phrases to the skill.
- **Batch-uniform assignment params on `greenlake_bulk_add_devices`** — `subscription_key` / `service_id` / `location` / `tags` apply to every device lacking that column (per-row CSV column wins). This is what lets an **uploaded** serial/MAC list — which the AI can't edit, because it's read server-side per the v3.4.6.x security model — be subscribed + service-assigned in one call.

## Notes
- Honors the upload-security model (server-side reads, no `read_file`) and serial tokenization from v3.4.6.1–v3.4.6.3.
- `bulk_add.py` stays under the 500-line limit: the tool-description constant, the uniform-assignment helper, and the serial-redactor factory were moved to `_bulk_source.py`.
- No change to the 3178 tool count (new params on an existing tool; a skill is not a platform tool).

## Validation
- `test_uploads_server_side`: uniform-assignment fills missing fields, per-row column wins, all-None leaves rows untouched.
- `test_skill_tool_references` / `test_skills`: the new skill's frontmatter + tool references validate.
- Full gate green: ruff + format + mypy (682 files) + **1965 passed / 1 skipped**.

Minor bump 3.4.6.3 → 3.4.7.0 (uv.lock re-locked).


nowireless4u:
Fixed. `_resolve_application_id` now tries an exact `item["id"]` match first (the service UUID an operator copies from the catalog), then falls back to the existing case-insensitive name/displayName/description substring match — so both forms work. Updated the `service_id` param description and the skill's Stage 3 to say "service ID **or** name; both resolve," and added `test_exact_id_match` covering the UUID path. Full gate green (1966 passed / 1 skipped).

PR #553, opened 2026-06-30
feat(greenlake): vendor public OpenAPI specs + auto-refresh workflow

nowireless4u:
## What
Vendors the **33 public northbound HPE GreenLake OpenAPI specs** (25 services) under `vendor/greenlake/`, driven by a curated `sources.json` manifest, plus a GitHub Actions workflow (`greenlake-specs-refresh.yml`) that re-pulls them weekly and opens a PR on change.

This is the foundation for **spec-driven GreenLake tool generation** (replacing the current hand-curated ~11-tool GreenLake surface) and the device-onboarding / subscription-matching work.

## How the specs were found
The portal is a Redocly SPA with no `sitemap.xml`/`redocly.yaml`/public repo. The reliable index is **`/llms.txt`** (the AI-crawler manifest). Raw specs are the `_bundle/...json?download` files — which **preserve source casing** (`openApi.json`, `openapiBeta.json`, `@v1`) while `llms.txt` paths are lowercased, so URLs **can't be string-derived** and are pinned (verified) in `sources.json`. The workflow diffs `llms.txt` on each run to flag newly-published services.

## Curation (one public northbound spec per service)
Excluded: `*-internal`, `*uiapi`/`*appapi`, events-only fragments, `changelog`, pure `-latest` duplicates. Notes:
- `audit-logs` collapsed to its primary API (it exposed 6 overlapping fragments).
- **`service-catalog` kept 5 distinct sub-APIs** (Service Manager Mgmt, Service Provision ×2, Service Registry ×2) — flag if you'd rather trim to just Service Manager Mgmt (the one onboarding uses).
- `identity` + `workspace` both vendored (same-titled specs but distinct services).
- Changelog-only/"coming soon" services (billing, caas, metal, mlops, iam/orgs, etc.) are absent — the workflow will flag them when HPE publishes specs.

## Testing
- Ran the workflow's **download step** and **drift-detection step** verbatim (extracted from the YAML) on the host: all 33 specs pull (HTTP 200 + valid OpenAPI), drift reports "No new services".
- Specs are canonicalized (`indent=2`, `sort_keys`) so future refreshes diff cleanly.
- **Actual `workflow_dispatch` run happens post-merge** — GitHub only allows dispatching a workflow once it's on the default branch.

## Scope / safety
- No `src/` runtime changes, no new tools, no version bump — infra + vendored data only.
- **Not** copied into the Docker image (the Dockerfile only `COPY`s `src/`/`pyproject.toml`/`README.md`), so no image bloat (~9 MB of specs stay in-repo only).

See `vendor/greenlake/README.md` for the full rationale.


nowireless4u:
Fixed — good catch. Rewrote `greenlake-specs-refresh.yml` to mirror the existing OAS-sync playbook exactly: checkout + PR creation via `secrets.MIST_SYNC_PAT` (not `GITHUB_TOKEN`), `peter-evans/create-pull-request@v6`, auto-merge, issue-on-failure, and a `concurrency` group. The download/validate logic moved into `.github/scripts/fetch_greenlake_oas.py` (stdlib `urllib`, sorted-key canonical output, non-zero exit on failure → issue fires) alongside `fetch_aruba_oas.py`. Verified the script pulls all 33 specs cleanly + idempotently.

One side note: I used the existing `ci/cd` label for the failure issue — the `openapi-sync`/`central` labels referenced by `sync-new-central-oas.yml` and `sync-mist-openapi.yml` don't actually exist in the repo, so those workflows' failure-issue step would 422. Pre-existing + only on the failure path, so I left them alone here, but flagging it.

PR #614, opened 2026-07-13
fix(edgeconnect): transport errors as 502 + collapse concurrent 401 refreshes (#602, #603)

### LINKED ISSUE #602: EdgeConnect transport errors return status_code 0 instead of upstream 502
**Severity:** low
**Area:** EdgeConnect

## Finding
EdgeConnect `format_http_error()` returns `status_code: 0` for non-HTTP exceptions such as DNS failures, TLS errors, and timeouts.

## Evidence
`client.py:359-374` returns `{"status_code": 0}` for non-`HTTPStatusError`. Probe with `httpx.ConnectTimeout("boom")` returned status 0.

## Impact
AI clients and middleware cannot classify EdgeConnect transport failures using the documented 4xx-vs-5xx status domain.

## Acceptance criteria
- Non-HTTP upstream/transport failures return a 502-style ToolError.
- HTTP status errors preserve actual upstream status/body.
- Tests cover timeout/connect error formatting.

### LINKED ISSUE #603: EdgeConnect concurrent 401 refresh can race by clearing shared session cookies
**Severity:** low
**Area:** EdgeConnect

## Finding
In user/password mode, concurrent 401 refreshes can clear the shared cookie jar under in-flight retries.

## Evidence
`client.py:160-162` clears `self._http.cookies` before login; `client.py:266-271` refreshes once on 401 and retries using the shared client/cookie jar. Mechanism is source-verified; no live flaky 401 reproduced.

## Impact
Concurrent EdgeConnect calls during session expiry can produce intermittent spurious 401 ToolErrors despite valid credentials, plus login bursts.

## Acceptance criteria
- Concurrent 401 refreshes collapse to one effective login or avoid clearing cookies needed by sibling retries.
- Retry requests cannot be built with an empty/stale cookie jar after sibling refresh.
- Concurrency test covers simultaneous 401s in user/password mode.

nowireless4u:
Cluster D of Casey Jones's full-project review — the two EdgeConnect Orchestrator client findings.

## Fixes

**#602 (low) — transport failures return a 502, not an unclassifiable `0`.** `format_http_error` mapped every non-`HTTPStatusError` exception (DNS failure, TLS error, connect/read timeout, connection reset) to `status_code: 0`, so AI clients and middleware couldn't place EdgeConnect transport failures in the documented 4xx-vs-5xx status domain. Non-HTTP upstream/transport failures now map to **502**; genuine HTTP status errors still preserve the real upstream status and body.

**#603 (low) — concurrent 401 refreshes collapse to one login.** In user/password mode a 401 forced an *unconditional* re-login, so two in-flight calls hitting session expiry each cleared and re-established the **shared** cookie jar — a sibling retry could be built against a jar another refresh had just cleared, producing intermittent spurious 401s and login bursts.

`AsyncTokenManager` (shared by all platforms) now carries a monotonic acquisition **`generation`**. The EdgeConnect client reads `(token, generation)` together via `get_token_with_generation()` and, on a 401, calls `refresh_if_stale(generation)` — which re-acquires **only if no sibling has already refreshed** since that generation. Concurrent victims serialize on the existing lock and collapse to a single login, so no retry is ever built against a jar a sibling just cleared. `get_token()` / `refresh()` behavior is unchanged.

## Acceptance criteria
- #602: non-HTTP transport failures → 502 ToolError ✅; HTTP status errors preserve status/body ✅; timeout/connect formatting covered by tests ✅.
- #603: concurrent 401 refreshes collapse to one effective login ✅; retries cannot be built against a jar a sibling cleared (single serialized login) ✅; concurrency test covers simultaneous 401s in user/pass mode ✅.

## Testing
- **Manager-level (`test_common_auth.py`)**: generation advances per acquisition; `get_token_with_generation` consistency; `refresh_if_stale` collapses two concurrent victims to exactly one re-fetch; still re-acquires when generation matches; skips when already advanced.
- **Client-level (`test_edgeconnect_client.py`)**: transport errors (`ConnectTimeout`/`ReadTimeout`/`ConnectError`/generic) → 502; HTTP status error preserves status+JSON body; `edgeconnect_request` wraps a transport error as a 502 ToolError; a barrier-synchronized two-request concurrency test proves simultaneous 401s produce exactly one re-login (would be 3 without the collapse).
- Full gate: ruff / format / mypy clean; **2032 passed**, 2 skipped.

Closes #602
Closes #603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.5.3.5] - 2026-07-13
+
+**Patch — EdgeConnect transport & auth robustness (Casey Jones full-project review: #602, #603).** Two low-severity correctness fixes to the EdgeConnect Orchestrator client.
+
+### Fixed
+- **#602 — transport failures return a 502, not an unclassifiable `0`.** `format_http_error` mapped every non-`HTTPStatusError` exception (DNS failure, TLS error, connect/read timeout, connection reset) to `status_code: 0`, so AI clients and middleware couldn't place EdgeConnect transport failures in the documented 4xx-vs-5xx status domain. Non-HTTP upstream/transport failures now map to `502`; genuine HTTP status errors still preserve the real upstream status and body.
+- **#603 — concurrent 401 refreshes collapse to one login.** In user/password mode a 401 forced an unconditional re-login, so two in-flight calls hitting session expiry each cleared and re-established the shared cookie jar — a sibling retry could be built against a jar another refresh had just cleared, producing intermittent spurious 401s and login bursts. `AsyncTokenManager` now carries a monotonic acquisition `generation`; the client reads `(token, generation)` together (`get_token_with_generation`) and, on a 401, calls `refresh_if_stale(generation)`, which re-acquires only if no sibling has already refreshed. Concurrent victims collapse to a single login serialized under the existing lock, so no retry is built against a jar a sibling cleared.
+
+### Changed
+- `AsyncTokenManager` (shared across all platforms) gains a `generation` property, `get_token_with_generation()`, and `refresh_if_stale(observed_generation)`. Existing `get_token()`/`refresh()` behavior is unchanged.
+
 ## [3.5.3.4] - 2026-07-13
 
 **Patch — ClearPass policy-visualizer correctness, part 2 (Casey Jones full-project review: #595, #598).** The simulation / graph-semantics fixes of the ClearPass visualizer cluster.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "hpe-networking-mcp"
-version = "3.5.3.4"
+version = "3.5.3.5"
 description = "Unofficial, community-driven MCP server for Juniper Mist, Aruba Central, HPE GreenLake, and ClearPass. Not affiliated with HPE, Aruba, or Juniper."
 readme = "README.md"
 license = "MIT"
--- a/src/hpe_networking_mcp/platforms/_common/auth.py
+++ b/src/hpe_networking_mcp/platforms/_common/auth.py
@@ -99,6 +99,7 @@ class AsyncTokenManager:
         self._buffer = expiry_buffer
         self._token: str | None = None
         self._expires_at: float | None = None  # Unix timestamp; None = no expiry
+        self._generation = 0  # bumped on every successful acquisition (see refresh_if_stale)
         self._lock = asyncio.Lock()
 
     @classmethod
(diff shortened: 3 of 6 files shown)


nowireless4u:
Thanks for the close look @casey-jones-ops — I checked this carefully and the blocking finding is a **false positive**: the return is already at the scope your suggested fix recommends.

In `get_token_with_generation()` the `return self._token, self._generation` sits at the `async with self._lock:` body level (12 spaces) — **after** the `if not self._is_fresh(): await self._fetch_locked()` block, not inside it:

```python
        async with self._lock:
            # Re-check inside the lock — another coroutine may have fetched.
            if not self._is_fresh():
                await self._fetch_locked()
            if self._token is None:                 # ← locked re-check guard
                raise AuthError(...)
            return self._token, self._generation    # ← reached whether WE fetched or a sibling did
```

So in your step-4/5 trace, coroutine B enters the lock, `if not self._is_fresh()` is now False (A already fetched), it **skips** the fetch block, falls through the `None` guard, and returns `(self._token, self._generation)`. There is no implicit `None` fall-through — that matches the `async with … return self._token, self._generation` snippet you proposed.

This was already covered by the pre-existing `test_concurrent_get_token_fetches_once` (5 concurrent stale callers through `get_token()` → `get_token_with_generation()`; a `None` return would have raised `TypeError` on the unpack). To address your request directly I also added an explicit regression (`fecd511`):

```python
async def test_concurrent_get_token_with_generation_from_stale_cache(self):
    # slow_fetch holds the lock via asyncio.sleep so 4 callers pile up behind it
    results = await asyncio.gather(*(mgr.get_token_with_generation() for _ in range(4)))
    assert len(calls) == 1                        # one fetch
    assert all(r == ("tok-1", 1) for r in results)  # no None fall-through; same (token, gen)
```

Green: `7 passed` for the auth concurrency set, full gate `2032 passed`. Ready for another pass when you get a chance.

nowireless4u:
I appreciate the diligence @casey-jones-ops, but this finding is incorrect, and I have two independent proofs. The code you quoted is not what is on the branch — your snippet indents `return self._token, self._generation` **inside** the `if not self._is_fresh():` block; the actual file has it one level out, in the `async with` body.

**Proof 1 — exact bytes at the reviewed commit `fecd511`** (leading-space count per line, `git show HEAD:…/auth.py`):

```
185 | indent= 8 |         async with self._lock:
186 | indent=12 |             # Re-check inside the lock — another coroutine may have fetched.
187 | indent=12 |             if not self._is_fresh():
188 | indent=16 |                 await self._fetch_locked()
189 | indent=12 |             if self._token is None:
190 | indent=16 |                 raise AuthError(...)
191 | indent=12 |             return self._token, self._generation
```

Line 191 (`return`) is at **indent 12** — identical to line 187 (`if not self._is_fresh()`) and line 189 (the `None` guard), while the fetch call on line 188 is at indent 16. So the `return` is **not** inside the fetch block; the lock-loser skips the fetch and still reaches it. This is byte-for-byte the structure your own suggested fix recommends.

**Proof 2 — CI Tests are green on this exact commit.** You wrote the new regression "should fail… so either CI has not reached this yet, or the claimed full gate was from a different local state." CI has now reached it: **Tests: pass** on `fecd511` (run 29337275554), along with Lint/Format, Type Check, and the rest. The `gather` of 4 stale callers returns `("tok-1", 1)` four times — no `None` fall-through.

The latch is shut. There is no code change to make here; making the `return` "return the cached pair after the locked re-check" is exactly what it already does. Please re-run against `fecd511` rather than the quoted snippet.

PR #619, opened 2026-07-15
feat: switch to .env file credentials with Docker secrets overlay

michaelrosejr:
## Summary

Replaces the mandatory Docker secrets setup with a simpler two-tier credential system:

1. **Environment variables** (recommended for dev) — set via `.env` file or shell export
2. **Docker secrets** (production) — unchanged, now via optional `docker-compose.secrets.yml` overlay

### Changes

- **`.env.example`** — new template; copy to `.env-<name>` and fill in credentials for platforms you use
- **`docker-compose.yml`** — removes mandatory `secrets:` block; credentials come from env vars. Adds `restart: unless-stopped` + `/livez` healthcheck
- **`docker-compose.secrets.yml`** — new optional overlay for production Docker secrets deployments
- **`config.py`** — `_read_secret()` now falls back to `NAME.upper()` env var after checking the secrets file. Adds `CUSTOMER_NAME` support
- **`run.sh`** / **`build.sh`** — added usage/argument validation guards
- **`README.md`** — updated setup instructions to reflect the simpler env file approach
- **`.gitignore`** — added `docs/env_secrets_migration_plan.md`, `AGENTS.md`, `docker-compose.override.yml copy.example`, `.rest`
- **`tests/unit/test_config.py`** — new tests for the env var fallback path

### Migration

Existing deployments using Docker secrets continue to work unchanged — the secrets tier still takes priority. No action required unless you want to switch to the simpler `.env` approach.


nowireless4u:
Thanks for this @michaelrosejr — nice work on the file-secret-priority + env fallback. Pushed maintainer fixes for @casey-jones-ops's three blockers and brought the branch up to date with `main`:

1. **Default bind restored to `127.0.0.1`.** Reverted the ports mapping to `127.0.0.1:${MCP_PORT:-8000}:8000` and added a comment on how to intentionally opt into `0.0.0.0` (only behind an authenticating reverse proxy / firewall). The in-container `MCP_HOST=0.0.0.0` is left as-is — host exposure is governed by the ports mapping, not the internal bind.
2. **EdgeConnect re-added** to `docker-compose.yml`, `.env.example`, and `docker-compose.secrets.yml` (host + api_key/user+password + login_type + verify_ssl). Also added the `ENABLE_GREENLAKE_WRITE_TOOLS` / `ENABLE_EDGECONNECT_WRITE_TOOLS` flags that were missing from `.env.example` (the compose already had them), so the two files match.
3. **Blank `ALLOWED_ORIGINS` now falls back to the localhost defaults** in `load_config()` (`not env or not env.strip()`), matching the code comment and Compose's `${ALLOWED_ORIGINS:-}` empty-string injection. Added `TestAllowedOrigins` covering blank / absent / explicit.

Full suite **2069 passed**; ruff / format / mypy clean. Ready for another pass.

nowireless4u:
```
gh pr diff 619 --repo nowireless4u/hpe-networking-mcp | python3 -c 'import sys,re\ntext=sys.stdin.read()\nfor path in ["docker-compose.yml","docker-compose.secrets.yml",".env.example","src/hpe_network...
```
Reason: Security scan — [HIGH] Pipe to interpreter: gh | python3: Command pipes output from 'gh' directly to interpreter 'python3'. Downloaded content will be executed without inspection.

Reply `/approve` to execute this one operation, `/approve session` to approve this pattern for the session, or `/deny` to cancel.

nowireless4u:
Fixed the backward-compat blocker @casey-jones-ops flagged.

**Restored the Docker-secret mounts to the base `docker-compose.yml`** (service `secrets:` list + top-level `secrets:` definitions, all platforms incl. EdgeConnect). So an existing deployment running the documented `docker compose up -d` gets its `/run/secrets/*` mounted exactly as before — no migration, no command change.

The `.env` support is preserved as an *additive* fallback: `_read_secret()` reads the secret file first (file wins), else the `NAME.upper()` env var — so file-secrets, `.env`, or a mix all work from the default `docker compose up -d`.

Since secrets are back in the base file, the separate **`docker-compose.secrets.yml` overlay was redundant, so I removed it** and updated the README (secrets setup + the "invalid mount config" troubleshooting note now point at `docker-compose.yml`) and the CHANGELOG entry to match. `docker compose config` validates; the compose still declares a secret file per platform, so the existing "comment out platforms you don't use" guidance still applies (also documented).

Net: existing Docker-secrets users are unbroken, `.env` users get the new path, default bind is loopback, EdgeConnect is wired in both credential paths. Ready for another pass.

nowireless4u:
Fixed the `.env`-only Compose path @casey-jones-ops flagged — thanks for the precise repro.

**Root cause** (as you found): the base `docker-compose.yml` mounts every declared Docker secret, so Docker fails container-create on the missing `secrets/*` files *before* `_read_secret()` can fall back to env vars.

**Fix — added a `docker-compose.env.yml` overlay** that clears the service secret mounts:

```yaml
services:
  hpe-networking-mcp:
    secrets: !reset []
```

Now each credential mode runs with **no manual file editing**:

- **Docker-secrets users (backward compatible, unchanged):**
  `docker compose up -d`
- **`.env`-only users:**
  `docker compose -f docker-compose.yml -f docker-compose.env.yml --env-file .env up -d`

I reproduced your exact failing scenario against the new head — env file with only Mist creds, **no** `secrets/*` files:

```bash
printf 'MIST_API_TOKEN=fake\nMIST_HOST=api.mist.com\n' > /tmp/env
docker compose --env-file /tmp/env -f docker-compose.yml -f docker-compose.env.yml \
  -p test up --no-start --no-build
# → Container ... Created ; exit=0   (no bind error)
```

`docker compose config -q` validates both the base and the overlaid graph.

**Docs updated to the two-path model:** `.env.example`, the README Quick Start / credential-methods / troubleshooting sections, and the CHANGELOG entry now show the overlay command for the `.env` path and the plain `docker compose up -d` for the secrets path. Existing Docker-secrets deployments remain untouched. Ready for another pass.

michaelrosejr:
All review feedback from @casey-jones-ops has been addressed. Ready for re-review.

PR #625, opened 2026-07-17
fix(deps): bump mcp 1.27.0 → 1.28.1 (clears 3 published CVEs)

nowireless4u:
Unblocks CI. **Not specific to any branch — `main` carries the same vulnerable pin**, so this currently fails every PR.

## Why now

`main`'s last Security run passed at 2026-07-16 19:23Z; #624's run failed at 2026-07-17 03:40Z. Nothing changed in the dependency between those times — the advisories were **published in between**. Re-running Security on `main` today would fail identically.

## The advisories

| CVE | Fixed in |
|---|---|
| CVE-2026-52869 | 1.27.2 |
| CVE-2026-52870 | 1.27.2 |
| CVE-2026-59950 | **1.28.1** |

1.28.1 (current latest) is the floor that clears all three.

## The change

- `mcp[cli]` floor raised **`>=1.20.0` → `>=1.28.1`**, not just re-locked — otherwise a future resolve could quietly fall back to a vulnerable release while the lockfile looked fine.
- `uv.lock` re-locked: `mcp 1.27.0 → 1.28.1`.
- Also refreshes a stale self-version in `uv.lock` (`3.4.7.1` → `3.6.0.1`) that had drifted from `pyproject.toml`.

## Verification

```
uv run --frozen pip-audit   -> No known vulnerabilities found
uv run --frozen pytest      -> 2085 passed, 2 skipped
```

No code changes were required for 1.28.1. The lock diff is narrow — only `mcp`, its constraint, and the self-version; no mass re-resolve.

Patch bump to **3.6.0.1** so the fix is identifiable and releasable on its own rather than waiting to ride along with a feature.

Kept separate from #624 deliberately: a security bump shouldn't be buried in a 28-file feature PR, and merging this fixes `main` for everyone. I'll rebase #624 on top once this lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 - **Bump `mcp` 1.27.0 → 1.28.1** to clear three advisories published against the locked version: CVE-2026-52869 and CVE-2026-52870 (fixed in 1.27.2) and CVE-2026-59950 (fixed in 1.28.1). The dependency floor is raised from `>=1.20.0` to `>=1.28.1` so a future resolve can't fall back to a vulnerable release. `pip-audit` reports no known vulnerabilities; full suite green on 1.28.1 with no code changes required.
-- Refreshes the stale `hpe-networking-mcp` self-version in `uv.lock` (3.4.7.1 → 3.6.0.0) as a side effect of re-locking.
+- Refreshes the stale `hpe-networking-mcp` self-version in `uv.lock` (3.4.7.1 → 3.6.0.1) as a side effect of re-locking.
 
 ## [3.6.0.0] - 2026-07-16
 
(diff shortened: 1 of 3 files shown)


nowireless4u:
Good catch @casey-jones-ops — fixed in `9979e21`.

You were exactly right about the cause: I wrote that line while the lock self-version was `3.6.0.0`, then bumped `pyproject.toml` to `3.6.0.1` and re-locked — which moved the self-version to `3.6.0.1` — and never went back to correct the note. The commit message had it right; only the changelog was stale. Agreed on the principle: release notes are the breadcrumb someone follows later, and a wrong version there is worse than no note.

The entry now reads `3.4.7.1 → 3.6.0.1`, matching the lockfile.

I also re-verified every other version claim in the entry against the actual files rather than just patching the one you flagged:

| claim | reality |
|---|---|
| heading `3.6.0.1` | `pyproject.toml` = `3.6.0.1` |
| `mcp` → `1.28.1` | `uv.lock` mcp = `1.28.1` |
| floor `>=1.28.1` | `pyproject.toml` = `>=1.28.1` |
| self-version → `3.6.0.1` | `uv.lock` = `3.6.0.1` |

All consistent now.

Issue #411, opened 2026-06-03
PII tokenization bypassed for show-command CLI output (output blob classified SKIP)


nowireless4u:
## Summary

`central_show_commands` (and other show-command tools that return raw device CLI output) embed the command output as a free-text string under the `output` field (`results[].output`). The PII walker classifies fields by name, and `output` / `command` / `results` are **not** in `FREE_TEXT_FIELD_NAMES` (`{description, notes, comment, comments, remarks, details}`), so the blob is classified **`SKIP`** and never regex-swept. Only `_universal_scan` (emails, PEM blocks, AWS-signed URLs) runs over it.

→ Any **tokenizable** PII embedded in CLI output — usernames, serials, hostnames, PSKs, RADIUS/SNMP/TACACS material that appears in `show run` / config / `show ucm` dumps — passes through **untokenized** when `ENABLE_PII_TOKENIZATION` is on (unless it happens to be an email/PEM/AWS-signed URL).

## Verified live (AOS-10 AP)

- `show ucm hashtable` → raw client MACs (`38:86:f7:…`) and client IPs in the `output` blob.
- `show ucm cdrs` → raw client IPs.

Note: MACs and IPs are *intentionally* never tokenized (MACs normalize-only; IPs preserved for audit), so those specific values are not themselves the leak — they demonstrate that the CLI blob is passed through unswept. The concern is **tokenizable** kinds (usernames, serials, secrets) that surface in other show commands.

## Impact

- Surface: any show-command tool returning CLI text — Central `central_show_commands`, plus AOS8 show-command equivalents.
- Only material when `ENABLE_PII_TOKENIZATION=true`.
- The device masks some secrets (RADIUS/TACACS shared secrets render as `********`), but PSKs come back cleartext — so a cleartext PSK in show output would leak.

## Suggested fix

- Classify the show-command output field(s) as `SCAN_FREE_TEXT` so the regex sweep runs over the blob, **or** add a dedicated CLI-text sweep step for show-command responses.
- Add a regression test asserting a known tokenizable token (e.g. a `username` / serial) embedded in a `show ...` output blob is tokenized end-to-end.

## How it surfaced

Found while live-validating feature request #400 (UCC/call-quality via `show ucm cdrs`).

nowireless4u:
Fixed in #463 (merged to main). Released in v3.3.10.1.

PR #626, opened 2026-07-17
fix(redaction,code-mode): device-serial leaks + hand-curated read-param gaps

nowireless4u:
Follow-up to #624. Closes the config-read parameter gaps in the hand-curated readers that the #624 sweep missed, and fixes three PII-tokenization defects plus one code-mode error-handling defect **found during testing** of those parameters.

All four were verified against a live tenant and are pinned by new unit tests.

## Security — PII tokenization

**1. Device serials leaked inside the `scope_device_function` annotation.**
Central returns `aruba-annotation:scope_device_function` as a *stringified* list of records (JSON or Python `repr`), so the redaction walker saw one opaque string and never descended to the identifiers inside it. A scope of type `DEVICE` is **named by the device's serial** — verified live, 119/119 DEVICE scope nodes had `scope_name == serial`. So every `detailed=True` read shipped that serial cleartext, while the *same* serial was tokenized under its own `serial` key elsewhere in the response — meaning the masking could be undone by correlating the two. This is the same bug class as #411 (structured data hiding in a string field). Fixed by parsing the blob and tokenizing DEVICE `scope_name` as `SERIAL` (deduped to the same token as the `serial` key); GLOBAL/SITE/DEVICE_COLLECTION labels stay cleartext per the existing privacy model.

**2. Known cleartext embedded in a larger string was not re-masked.**
The keymap replay only matched a value that was a known plaintext *in its entirety*, so a tokenized identifier stayed cleartext the moment anything was concatenated around it (e.g. `management-users/<user>`). Added `Tokenizer.replace_known_cleartext` — word-bounded, length-floored embedded substitution so it can't corrupt unrelated text.

**3. Certificate fields over-masked non-material scalars.**
`TokenKind.CERT` fired on any value under a `cert`/`certificate`/`chain`/… key regardless of shape, so a certificate *name* — or a caller's own scalar reusing the word — became a `[[CERT:…]]` token. Because tokens are keyed by plaintext, identical throwaway values collided onto one token and read like a tokenizer fault (the tell that surfaced this during testing). CERT now requires certificate material (PEM/OpenSSH armor, or a long opaque blob); every other exact-match secret kind still fires unconditionally.

## Fixed — code mode

**4. A bad tool argument killed the whole `execute()` block.**
FastMCP raises its *own* `fastmcp.exceptions.ValidationError` (a `FastMCPError`, **not** a subclass of pydantic's) for an unexpected keyword argument during binding. `ValidationCatchMiddleware` caught only pydantic's type, so this escaped into the sandbox as a fatal error — a model probing a tool with a wrong kwarg lost its entire `execute()` block even inside `try/except`. The middleware now catches both, recovers the structured pydantic error from `__cause__` for a clean redacted summary, and strips `input_value=` echoes from the string fallback so a mis-supplied secret can't leak. Returns the normal `ok=False, status=422` envelope the model can branch on.

## Read-param gaps (the #624 sweep keyed on `_get_resource` callers)

- **`central_get_wlan_profiles`** — had none of the eight query params; now delegates to `_get_resource` and inherits the scope-guard + pagination handling.
- **`central_get_role_with_policy`** — gains the six meaningful scope params, applied to **both** the role and each bound-policy fetch so the bundle is one consistent view. No `limit`/`offset` — it resolves named objects.
- **`central_get_config_assignments`** — gains the documented `profile_type` filter, and its `scope_id`/`device_function` (which had `Field(default=None)` but no Python default) can now actually be omitted, matching the docstring (a direct call previously raised `TypeError`).

## Verification

- **Live, end-to-end:** ran the real tokenizer over `detailed=True` reads across five config types; **0 of 119** device serials leaked cleartext (was leaking on every `scope_device_function` blob before).
- **Live:** reproduced the sandbox death exactly (a bad `object_type` kwarg), confirmed the `execute()` block now survives with a clean `422` envelope and no value echo.
- **Tests:** +29 unit tests (`test_redaction_embedded_leaks.py`, extended `test_validation_catch.py`). Full suite **2134 passed**; ruff / format / mypy / bandit clean.

Refs #624


nowireless4u:
Fixed in `a793e4d` — good catch, that was a real leak in the exact path meant to prevent one.

You were right: `_INPUT_ECHO_RE = input_(?:value|type)=[^,\]]+` stopped at the first comma or `]`, so `input_value='secret,still-secret'` left `still-secret'` in the message. Confirmed before fixing:

```
input_value='secret,still-secret', input_type=str]
  old regex -> "...[type=unexpected_keyword_argument,still-secret']"   # leaks
```

The value is a Python repr that can hold commas, quotes, and brackets, so no single-delimiter bound is safe. Fixed by anchoring on the **line end** instead — pydantic renders each error's `[type=..., input_value=..., input_type=...]` on one line and escapes any real newline in the value to a literal `\n`, so stripping `input_value=` to end-of-line removes the value, `input_type`, and any URL in one shot while keeping the field name + human message on the surrounding lines.

Added parametrized regression tests for the shapes you named — value with a comma, with `]`, with both, and a quoted value — each asserting the secret and `input_value` are gone and the field name survives. Full suite 2138 passed; ruff/format/mypy/bandit clean.

PR #630, opened 2026-07-18
feat(mist): resync tool surface from OpenAPI (+13 tools, +5 param fixes)

nowireless4u:
Second of two PRs from the cross-platform audit. The Mist audit flagged 5 tools missing spec-documented query params — but Mist is **generated**, so the fix is a resync, not a hand-patch. Per the compare you asked for, this is safe.

## The compare (current tools vs. synced OAS)

| | Count |
|---|---|
| OAS operations | 1050 |
| Current tools | 1037 |
| **ADD** | **13** (new endpoints) |
| **REMOVE** | **0** |
| **Param drift** | **5** |

**Coverage-additive — nothing removed.** Every current Mist tool still maps to a live spec operation.

## What changed
- **+13 tools**: Marvis client stats/events/metrics (`mist_*_marvisclients_*`), org device claims, sitemap `marvis-configs` — endpoints added upstream since the last generation.
- **5 param-gap fixes**: `map_id` (stats/assets), `mapstack_id` (maps), `disconnected_before` (inventory), `end_of_sale_before`/`_after` (JSI search), SCEP client-cert filters.
- Minor spec-driven description refreshes on a few tools.

## Why the diff looked terrifying (and isn't)
The raw regen was **210 files / −11,764 / +5,062**. That was **100% formatting** — the generator emits single-quotes and unwrapped annotations that `ruff format` normalizes right back. After formatting, the **true diff is 15 files, +542/−235** (13 modified + 2 new). `admins.py`, representative of ~197 "changed" files, had zero real change.

## Verification
Regenerated via `scripts/internal/regenerate_mist_tools.py`; **1050 tools register**; full suite **2154 passed**; ruff / format / mypy clean. Mist count updated across README / TOOLS / INSTRUCTIONS (1037 → 1050; total → 4149).

> **Merge order:** versioned **3.6.1.3**, assuming #629 (3.6.1.2) merges first. If this lands first I'll rebump #629.


nowireless4u:
Fixed both in \`8a4333f\`.

**CI failure (root cause the checks were dying at ~8s):** \`uv.lock\` carried an unresolved rebase conflict marker — the self-version block had \`<<<<<<< HEAD ... 3.6.1.2 / 3.6.1.3 ...\`. My in-rebase \`uv lock\` didn't rewrite it because uv refuses to parse a lockfile with conflict markers, and my local gate passed only because the dev container reused its existing venv instead of re-parsing the lock. CI's \`uv sync --frozen\` hit it immediately, which is why Lint/Tests/Type/Bandit/Audit all failed in seconds. Resolved to 3.6.1.3; \`uv lock --check\` is clean and \`uv sync --frozen\` now installs \`hpe-networking-mcp==3.6.1.3\`.

**Stale counts — you were right, my sed missed several.** Now updated:
- \`README.md\` comparison-table row (\`1037 + 2 prompts\` → \`1050 + 2 prompts\`) and the architecture diagram box (\`1037\` → \`1050\`).
- \`docs/TOOLS.md\` — the tag-bucket line (\`mist (1037 tools)\` → \`1050\`) and all five \`4136\` → \`4149\`.
- \`INSTRUCTIONS.md\` — the remaining \`4136\` → \`4149\`.

No \`1037\` or \`4136\` references remain in any of the three files. (The \`1216\` you noted on that row is EdgeConnect's own column, correct as-is.)

Full suite **2154 passed**; \`uv sync --frozen\`, ruff, format, mypy all clean. Thanks for catching both.

